PR #27, opened 2021-06-29
PUD-90: Run & monitor end-to-end tests on master after dev deploy and before preprod-approval

philsegal:


.circleci/config.yml
```diff
@@ -32,6 +32,55 @@ jobs:
           path: build/test-results
       - store_artifacts:
           path: build/reports/tests
+  end-to-end-tests:
+    executor:
+      name: hmpps/java
```
dazoakley, line 37: Any particular reason to use a java image/executor here?  I'm assuming it's a pretty hefty machine image.

You could probably get away with `hmpps/default_small` as you're just running bash scripts.

.circleci/config.yml
```diff
@@ -32,6 +32,55 @@ jobs:
           path: build/test-results
       - store_artifacts:
           path: build/reports/tests
+  end-to-end-tests:
+    executor:
+      name: hmpps/default_small
+      tag: "16.0"
```
dazoakley, line 38: Remove the tag line :)

dazoakley (approved):
LGTM 👍

PR #34, opened 2021-07-01
NJ Remove blocking in controller

mr-ken:


mikedunfordEE (approved):
Approving.   But not too keen the reactor-core / Mono stuff, could just be I am unfamiliar with this library but doesn't seem very nice at first glance (the test definitely is very bloated for a simple assertion).  Perhaps we need to discuss as team it's benefits and if we want to use it?

PR #35, opened 2021-07-02
PUD-191: Book recall endpoint

philsegal:


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallController.kt
```diff
@@ -0,0 +1,37 @@
...
+import java.util.UUID
+
+@RestController
+@RequestMapping(produces = [MediaType.APPLICATION_JSON_VALUE])
+@PreAuthorize("hasRole('ROLE_MANAGE_RECALLS')")
+class RecallController(
+  @Autowired private val recallRepository: RecallRepository
+) {
+
+  @PostMapping("/recall")
```
mr-ken, line 24:
```suggestion
  @PostMapping("/recalls")
```
I think using the plural in the API is a bit nicer because it means we expect a list of recall. And we can have `/recalls/{recall_id}` when getting a single recall.
If we look at other APIs in HMPPS Digital, they've also adopted the plural form for resources.

philsegal: Feel like we should agree this as a team. If we decide to go with plurals then we can make sure everything is consistent? We went with singular for the database? And if we rename the endpoint, then should probably rename the repository and controller also.

mr-ken: I think if we don't use plurals for our API we might get some comments from Digital that we're not following the canonical form (I had a brief look at their APIs and all of them use the plural form).
For the database and the code, I'm less concerned but I'm used to using the singular for relational DB tables.

jonnywyatt: I agree, plural is better here if it can return a list

philsegal: Well, this post won't return a list. but the next story will.. Plurals it is...

mr-ken:
It looks like the path is still singular (`/recall`) ; did you forget to push something?

philsegal:
3rd time lucky 🤦‍♂️ . Damn force-push craziness.

mr-ken (approved):
LGTM

PR #36, opened 2021-07-02
PUD-96 Find all recalls

mr-ken:


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallsController.kt
```diff
@@ -22,16 +22,19 @@ class RecallsController(
...
 
   @PostMapping("/recalls")
-  @ResponseBody
   fun bookRecall(@RequestBody bookRecallRequest: BookRecallRequest) =
     ResponseEntity(
       recallRepository.save(bookRecallRequest.toRecall()).toResponse(), HttpStatus.CREATED
     )
+
+  @GetMapping("/recalls")
+  fun findAll(): MutableList<Recall> = recallRepository.findAll()
```
philsegal, line 31: Personally, would return the `BookRecallResponse` object, rather than exposing our domain objects. Maybe rename it to `RecallResponse`?

mr-ken: yup - makes sense 👍🏻

build.gradle.kts
```diff
@@ -38,6 +38,7 @@ dependencies {
   testImplementation("io.zonky.test:embedded-database-spring-test:2.0.1")
   testImplementation("io.zonky.test:embedded-postgres:1.3.0")
   testImplementation("io.projectreactor:reactor-test")
+  testImplementation("com.ninja-squad:springmockk:3.0.1")
```
philsegal, line 41: Do we really want another mocking framework? We have the Zonky stuff that would work for integration tests for DB too?

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/integration/RecallsIntegrationTest.kt
```diff
@@ -0,0 +1,35 @@
+package uk.gov.justice.digital.hmpps.managerecallsapi.integration
+
+import com.ninjasquad.springmockk.MockkBean
+import io.mockk.every
+import org.junit.jupiter.api.Test
+import uk.gov.justice.digital.hmpps.managerecallsapi.db.Recall
+import uk.gov.justice.digital.hmpps.managerecallsapi.db.RecallRepository
+import java.util.UUID
+
+class RecallsIntegrationTest : IntegrationTestBase() {
```
philsegal, line 10: Should this be merged with the `BookRecallsIntegrationTest`? I assume you were writing this whilst I was writing the other, but nicer to have it clean and tidy.

build.gradle.kts
mr-ken: It's not another mocking framework - it's a Spring extension for Mockk so that you can use the `@MockkBean` annotation in an integration test.

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/integration/RecallsIntegrationTest.kt
mr-ken: I'm not against it but I can see some arguments for having separate tests suites as well.

build.gradle.kts
mr-ken: The embedded database doesn't spare you from wiring in the repository. This means that in your test you would still have to write recalls to the database if you want to get the recalls. If you wire in a mock repository, like you do with `@MockkBean`, you don't have to write as much boiler plate to set up your data and also you can test some edge unhappy paths a bit easier, i.e. repository throws some weird exception.

mikedunfordEE: From my point of view I try to avoid using 'lateinit' as much as possible, always use a 'val' if possible, so anything that avoids it works for me :-) This style is always my preference:

`private val recallRepository = mockk<RecallRepository>()`

But I know some people like annotations, so not too fussed either way.

mr-ken: I don't think you can use that code here. The `MockkBean` annotation also tells Spring that the RecallRepository bean in the context of this test should be instantiated as a mock.

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallsController.kt
```diff
@@ -22,16 +22,19 @@ class RecallsController(
 ) {
 
   @PostMapping("/recalls")
-  @ResponseBody
```
philsegal, line 25: How come we dont need this?

philsegal (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallsController.kt
mr-ken: it's redundant (says so in the javadoc)

PR #40, opened 2021-07-07
Add validation for /search endpoint, must provide a non-blank nomsNumber

mikedunfordEE:


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/config/ManageRecallsApiExceptionHandler.kt
```diff
@@ -7,72 +7,57 @@ import org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR
...
     return ResponseEntity
       .status(BAD_REQUEST)
-      .body(
-        ErrorResponse(
-          status = BAD_REQUEST,
-          userMessage = "Validation failure: ${e.message}",
-          developerMessage = e.message
-        )
-      )
+      .body(ErrorResponse(BAD_REQUEST, "Validation failure: ${e.message}"))
```
mr-ken, line 26: curious why you removed the developer message?

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/GenerateRevocationOrderController.kt
```diff
@@ -19,9 +18,7 @@ import java.util.Base64
 class GenerateRevocationOrderController(
   @Autowired private val pdfDocumentGenerator: PdfDocumentGenerator
 ) {
-  companion object {
-    val log: Logger = LoggerFactory.getLogger(this::class.java)
-  }
+  private val log = LoggerFactory.getLogger(this::class.java)
```
mr-ken, line 21: not a biggie but I think the reason why it was in a companion object is so that the log field is static.

mr-ken (approved):
Looks good - just a couple of comments. Agree that we need to have a chat about common code format standard.

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/config/ManageRecallsApiExceptionHandler.kt
mikedunfordEE: This was inherited from the template.  My use of this class is the first real use that I know of, so I simplified things.  The UI, as our only client, certainly doesn't use any of these error details yet.

mr-ken: ok thanks - we might find out later what's the purpose of that message...

PR #41, opened 2021-07-08
PUD-217: API: Add gender etc. to offender properties returned by search

AndyNewton:
PUD-217: API: Add gender etc. to offender properties returned by prisoner /search


mikedunfordEE (approved):
Lovely jubbly.  Makes me think we need to add some tiny types for everything instead of just strings everywhere.  Worth discussing first in case anyone disagrees

mr-ken (approved):
LGTM

PR #42, opened 2021-07-09
Add Pact provider test and circleCI workflow

mikedunfordEE:


src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/integration/IntegrationTestBase.kt
```diff
@@ -21,6 +23,7 @@ import java.util.concurrent.ThreadLocalRandom
 @SpringBootTest(webEnvironment = RANDOM_PORT)
 @ActiveProfiles("test")
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@AutoConfigureEmbeddedDatabase(provider = ZONKY)
```
mr-ken, line 26: We don't need an embedded database for every integration test (e.g. `RecallsIntegrationTest`) so it would be preferable if annotation was only added for the appropriate tests.

mr-ken:
Is my understanding correct that the pact tests will run in a separate workflow from "build and deploy" and that, as such, automatic deployment to the DEV env won't be impacted by a failure in the pact tests?

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/integration/IntegrationTestBase.kt
mikedunfordEE: Not sure I agree with integration tests that mock out the database.  But I can remove the annotation.  Maybe we need a different TestBase then for ones that use a DB or not

mikedunfordEE:
> Is my understanding correct that the pact tests will run in a separate workflow from "build and deploy" and that, as such, automatic deployment to the DEV env won't be impacted by a failure in the pact tests?

The pact tests still run as part of the build and deploy step.  The extra workflow is so we can trigger it independently from the pact broker and it can publish back to the broker the fact that it has verified the pact file.

This is all work in progress and trying to get the basics integrating together.  We can eventually look at hooking things into the deploy pipeline in a better way so we don't deploy if there are incompatible versions etc.  Small steps.  Don't want this to be sat on a PR for ages while we figure out every nuance

mr-ken:
> > Is my understanding correct that the pact tests will run in a separate workflow from "build and deploy" and that, as such, automatic deployment to the DEV env won't be impacted by a failure in the pact tests?
> 
> The pact tests still run as part of the build and deploy step. The extra workflow is so we can trigger it independently from the pact broker and it can publish back to the broker the fact that it has verified the pact file.
> 
> This is all work in progress and trying to get the basics integrating together. We can eventually look at hooking things into the deploy pipeline in a better way so we don't deploy if there are incompatible versions etc. Small steps. Don't want this to be sat on a PR for ages while we figure out every nuance

No worries - I'm trying to understand how this fits together with CI and CD and I do understand that this is a work in progress. 

So if I understand correctly, if the pact tests fail then CI will be broken but we only publish to the pact server when the pact worklfow is triggered manually in CircleCI. Is that right?

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/integration/IntegrationTestBase.kt
mr-ken: `RecallsIntegrationTest` is specifically testing the REST interface. We have another integration test `RecallRepositoryIntegrationTest` which tests the persistence layer using an embedded Postgres database. The advantage is that each of these tests focuses solely on one integration aspect. Also, `RecallsIntegrationTest` runs faster if it doesn't have to spin up a DB and it means you can test out scenario with the mock like if an exception is thrown by the repository and you expect a 500 response with a particular message - that's more tricky to do with an embedded database.

mr-ken (approved):

PR #57, opened 2021-07-20
PUD-106 Ability to persist recall documents

mr-ken:


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/db/RecallDocument.kt
```diff
@@ -0,0 +1,23 @@
...
+data class RecallDocument(
+  @Id
+  val id: UUID,
+
+  @Column(name = "recall_id", nullable = false)
+  val recallId: UUID,
+
+  @Column(name = "category", nullable = false)
+  @Enumerated(EnumType.STRING)
+  val category: RecallDocumentCategory
```
philsegal, line 22: I think we said we were going to store file name too?

philsegal (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/db/RecallDocument.kt
mr-ken: I was thinking the file name could be `{category}.pdf` since there's only one category of document per recall - if necessary, we can always add it later.

PR #60, opened 2021-07-21
Tiny types

mikedunfordEE:


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/config/ManageRecallsApiExceptionHandler.kt
```diff
@@ -18,6 +19,7 @@ import javax.validation.ValidationException
 class ManageRecallsApiExceptionHandler {
   private val log = LoggerFactory.getLogger(this::class.java)
 
+  //TODO:  Is this needed?
```
philsegal, line 22: Should we be diligent and only add `TODO` with a Jira number?

philsegal (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/config/ManageRecallsApiExceptionHandler.kt
mikedunfordEE: Adding a jira task might be overkill in my opinion.  Don't really want to have to add a jira ticket for every TODO, I would be less likely to add the TODO in the first place.

Obviously don't want the code littered with TODOs but if we are all diligent in trying to remove them it shouldn't be a problem.

In this case I added it just in case anyone knows of a reason they should be kept.  I don't think they should, but didn't want to add more to my PR as it was already quite large.  Hopefully I will delete them next PR and nothing will fail and they will go away :-)

mr-ken (commented):
It makes sense. I have to say that it feels like a lot of effort to keep maintaining these 2 tiny types and I'm not convinced we'll get that much benefit out of them.

PR #61, opened 2021-07-21
PUD-106 Ability to add documents to a recall

mr-ken:


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallDocumentService.kt
```diff
@@ -0,0 +1,36 @@
...
+import java.util.UUID
+import javax.persistence.EntityNotFoundException
+
+@Service
+class RecallDocumentService(
+  @Autowired private val s3Service: S3Service,
+  @Autowired private val recallRepository: RecallRepository
+) {
+
+  private val log = LoggerFactory.getLogger(this::class.java)
```
philsegal, line 19: this appears unused

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallDocumentService.kt
```diff
@@ -0,0 +1,36 @@
...
+
+  fun addDocumentToRecall(recallId: UUID, documentBytes: ByteArray, documentCategory: RecallDocumentCategory) {
+    val recall = try {
+      recallRepository.getById(recallId)
+    } catch (e: EntityNotFoundException) {
+      throw RecallNotFoundError("No recall found with ID '$recallId'", e)
```
mikedunfordEE, line 25: Could think about having a service level exception handler for EntityNotFoundExceptions, I'm guessing the service should always respond with 404 if this happens for any missing Entity.  Then we don't need specific exception logic for every  entity

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/storage/S3BulkResponseEntity.kt
```diff
@@ -2,10 +2,9 @@ package uk.gov.justice.digital.hmpps.managerecallsapi.storage
 
 import java.util.UUID
 
-data class S3BulkResponseEntity constructor(
+data class S3BulkResponseEntity(
```
mikedunfordEE, line 5: Was there before, but these should all be val not var

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallDocumentService.kt
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    s3Service.uploadFile(documentBytes)
```
mikedunfordEE, line 28: Need to use the UUID returned in the S3BulkResponseEntity when saving the document don't we?

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallDocumentService.kt
mr-ken: Isn't it better to throw a specific exception rather than a Hibernate one? I think it gives us more control. Then we can have a RecallNotFoundError handler which responds with a 404 if necessary? In this case I don't think you would respond with a 404 when you're creating a document. Probably just a 400, i.e. the operation that you are attempting is not valid since the recall doesn't exist.

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallDocumentService.kt
mr-ken: yes - good catch - thanks

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/storage/S3BulkResponseEntity.kt
mr-ken: Correct - I don't think we need this object tbh

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallDocumentService.kt
mikedunfordEE: We're kind of tied to Hibernate so I don't mind having a specific handler for that exception.  Plus it would then cater for any new hibernate types, otherwise we'll need to catch for every future repository and add a new Exception type.  Just thinking of ways to reduce the code and simplify.  We don't need to worry too much about it now as we only have one repository, just need to make sure this error is handled correctly

Something not found to me should be a 404.  A 400 signifies there's something invalid about the request you sent, it wasn't an invalid request, just the request was being made for something didn't exist

mr-ken: We're talking about code that doesn't exist but:
REST is about resources 
404 means "resource not found"
If we add an endpoint for `POST /recalls/{recallId}/documents` the resource is "documents"
In this case the document resource has not even been created so responding with 404 is does not make sense from a REST point of view, or at least would be very confusing

mikedunfordEE: Exactly

404 means "resource not found"

The documents can not be found, therefore 404.  Makes perfect sense to me, not sure what is confusing about it

mr-ken: It's not the document that hasn't been found, it's the recall ; the document doesn't exist yet.

mikedunfordEE: Ok, so we would be returning 404 because the Recall was not found, because it does not exist, same as the documents.

Under what circumstances would you return a 404?

mikedunfordEE (approved):

PR #62, opened 2021-07-21
PUD-106 Simplify S3Service

mr-ken:
- simpler interface upload(bytes) -> UUID, download(UUID) -> bytes
- removed some redundant code


mikedunfordEE (approved):
All good.

Although just spotted RecallNotFoundError extends Throwable, which it shouldn't, can this be RuntimeException or some other more applicable Spring exception type please?  Happy to fix this going forwards as it isn't being used yet

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/storage/S3StorageService.kt
```diff
@@ -18,30 +18,18 @@ class S3StorageService(
...
-      .also { log.info("uploaded $uuid to $bucketName: ${it.statusCode()}") }
-      .let { response ->
-        result = S3BulkResponseEntity(
-          bucket = bucketName,
-          fileKey = uuid,
-          successful = response.isSuccessful,
-          statusCode = response.statusCode()
-        )
-      }
-    return result
```
philsegal, line 45: Do we lose any error handling here?

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/storage/S3StorageService.kt
```diff
@@ -18,30 +18,18 @@ class S3StorageService(
...
+  override fun uploadFile(fileBytes: ByteArray): UUID {
+    val fileS3Key = UUID.randomUUID()
+    log.debug("Uploading file to s3://$bucketName/$fileS3Key")
+    val request = PutObjectRequest.builder().bucket(bucketName).key(fileS3Key.toString()).build()
+    val body = RequestBody.fromBytes(fileBytes)
+    s3Client.putObject(request, body)
```
philsegal, line 32: is there anything that is returned by putObject that we should be checking?

mr-ken:
> All good.
> 
> Although just spotted RecallNotFoundError extends Throwable, which it shouldn't, can this be RuntimeException or some other more applicable Spring exception type please? Happy to fix this going forwards as it isn't being used yet

Thanks for your review Mike. Out of curiosity, why shouldn't `RecallNotFoundError` extend Throwable?

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/storage/S3StorageService.kt
mr-ken: That code wasn't used at all. We can add it at a later stage if required.

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/storage/S3StorageService.kt
mr-ken: I had a look at the doc: `putObject()` throws different AWS specific exceptions in case of errors - so we don't need to check the internal HTTP status codes to confirm that the operation was successful ; we can just handle the exceptions. Perhaps we should encapsulate these exceptions with a specific exception like `UploadError`?

mr-ken: From the Javadoc:
```
Throws:
SdkClientException - If any errors are encountered in the client while making the request or handling the response.
AmazonServiceException - If any errors occurred in Amazon S3 while processing the request.
```

mikedunfordEE:
> > All good.
> > Although just spotted RecallNotFoundError extends Throwable, which it shouldn't, can this be RuntimeException or some other more applicable Spring exception type please? Happy to fix this going forwards as it isn't being used yet
> 
> Thanks for your review Mike. Out of curiosity, why shouldn't `RecallNotFoundError` extend Throwable?

Because there is never (or hardly ever) a good reason to extend Throwable.  Exception or RuntimeException is the lowest level in the hierarchy you should go.  We should try to be as specific in our type of exception as possible, especially as Spring has a plethora of exception types at our disposal.  I'm sure it's in several java books I've read over the years (although my ageing memory fails me as to which ones).  A quick google also throws up several other opinions on the subject

mr-ken:
> > > All good.
> > > Although just spotted RecallNotFoundError extends Throwable, which it shouldn't, can this be RuntimeException or some other more applicable Spring exception type please? Happy to fix this going forwards as it isn't being used yet
> > 
> > 
> > Thanks for your review Mike. Out of curiosity, why shouldn't `RecallNotFoundError` extend Throwable?
> 
> Because there is never (or hardly ever) a good reason to extend Throwable. Exception or RuntimeException is the lowest level in the hierarchy you should go. We should try to be as specific in our type of exception as possible, especially as Spring has a plethora of exception types at our disposal. I'm sure it's in several java books I've read over the years (although my ageing memory fails me as to which ones). A quick google also throws up several other opinions on the subject

I think you might be confusing the java Throwable (`java.lang.Throwable`), which I agree is not good practice to extend, with the Kotlin Throwable(`kotlin.Throwable`) which is the base class for all Exceptions in Kotlin (https://kotlinlang.org/docs/exceptions.html#try-is-an-expression). There are no checked exceptions in Kotlin so there's no need to extend `RuntimeExcpetion` like you would in Java to avoid having to catch checked exceptions.

PR #70, opened 2021-07-28
NJ Append console logs in Logstash format

mr-ken:
So that the logs can be indexed correctly in Elasticsearch and
thus can be filtered appropriately in Kibana


AndyNewton (approved):
Approving - albeit that I don't understand all the implications - we will change again if we need to of course.

PR #74, opened 2021-07-29
Add Update Recall endpoint and tests

mikedunfordEE:


AndyNewton (approved):
Mike has walked me through all of ^^^ and all LGTM.

mr-ken (approved):
Looks good. Out of curiosity, why not use PUT for an update instead of PATCH? I would say PUT is more conventional than PATCH.

mikedunfordEE:
> Looks good. Out of curiosity, why not use PUT for an update instead of PATCH? I would say PUT is more conventional than PATCH.

After discussion with Jon we decided on PATCH.  It could initially be a PUT, but we saw it as a generic 'update recall' endpoint going forwards.  So the UI would only supply the fields it wants updating and the backend would figure it out.  Happy to use either PUT or PATCH though

PR #78, opened 2021-08-04
Point at dev by default and account for user names with dots

dazoakley:


AndyNewton (approved):

AndyNewton:
Tested from local and working - as applied to clear down dev: https://mojdt.slack.com/archives/C020S8C0K9U/p1628093315090800?thread_ts=1628083079.088400&cid=C020S8C0K9U

PR #81, opened 2021-08-09
PUD-355 API - get and update (patch) recallEmailReceivedDateTime

AndyNewton:
PUD-355 API - get and update (patch) recallEmailReceivedDateTime


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallsController.kt
```diff
@@ -25,6 +25,7 @@ import uk.gov.justice.digital.hmpps.managerecallsapi.service.RecallDocumentServi
 import uk.gov.justice.digital.hmpps.managerecallsapi.service.RecallNotFoundException
 import uk.gov.justice.digital.hmpps.managerecallsapi.service.RevocationOrderService
 import java.net.URI
+import java.time.ZonedDateTime
```
philsegal, line 28: I assume this has all been discussed, and sorry for dragging it up again, but I assume we need a zone? I thought I saw some chat in slack around UTC?

AndyNewton: We agreed that the passed values, in the API, should include a TZ designator and the simplest thing then appears to be simply to include throughout the stack.  
https://dsdmoj.atlassian.net/browse/PUD-253?focusedCommentId=167611

philsegal (approved):

PR #83, opened 2021-08-11
Pud 365 add latest release place and date

NataliaOsk:


src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/UpdateRecallControllerTest.kt
```diff
@@ -168,4 +168,31 @@ class UpdateRecallControllerTest {
...
+  @Test
+  fun `can update recall with lastReleaseDateTime and lastReleasePrison without changing other properties`() {
+    val recallId = ::RecallId.random()
+    val priorRecall = Recall(recallId, nomsNumber, recallLength = FOURTEEN_DAYS, agreeWithRecallRecommendation = false)
+    every { recallRepository.getByRecallId(recallId) } returns priorRecall
+    val lastReleaseDateTime = ZonedDateTime.now()
+    val updatedRecall = priorRecall.copy(recallType = FIXED, lastReleasePrison = "prison", lastReleaseDateTime = lastReleaseDateTime)
+    every { recallRepository.save(updatedRecall) } returns updatedRecall
+
+    val response = underTest.updateRecall(recallId, UpdateRecallRequest(lastReleasePrison = "prison", lastReleaseDateTime = lastReleaseDateTime))
```
AndyNewton, line 181: For future: I would favour two changes here:
* extracting the value "prison" to a val and using that 3 times, rather than the literal 3 times
* superseding that fixed value ("prison") with variable data - some form of random string.
But they can come after.

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/integration/IntegrationTestBase.kt
```diff
@@ -128,6 +130,8 @@ abstract class IntegrationTestBase {
     recallType = RecallType.FIXED,
     agreeWithRecallRecommendation = agreeWithRecallRecommendation,
     recallLength = recallLength,
+    lastReleasePrison = "prison",
```
AndyNewton, line 133: I think this is an example of why we should use random or at least variable data for testing.
In fact I'm surprised that the IDE does not give a warning - that the `` parameter to the call is un-used.  There may be something I don't know going on with default values in Kotlin here.  
Certainly I think we're being inconsistent in this function now; I'll try to call you.

AndyNewton (changes requested):
Let's try to catch up on a call.

AndyNewton (approved):

PR #85, opened 2021-08-16
PUD-XX: Rename fields and change datetime to date.

philsegal:


mikedunfordEE (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/UpdateRecallController.kt
```diff
@@ -12,6 +12,7 @@ import org.springframework.web.bind.annotation.RestController
 import uk.gov.justice.digital.hmpps.managerecallsapi.controller.RecallType.FIXED
 import uk.gov.justice.digital.hmpps.managerecallsapi.db.RecallRepository
 import uk.gov.justice.digital.hmpps.managerecallsapi.domain.RecallId
+import java.time.LocalDate
```
jonnywyatt, line 15: Local to the server? Will this always be Europe/London?

philsegal: it has no timezone. it is timezone agnostic.

mikedunfordEE (approved):

PR #86, opened 2021-08-17
PUD-247: Add sentencing info fields

philsegal:
- cleanup unnecessary Column annotations


src/main/resources/db/migration/V14__Add_recall_sentencing_columns.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE recall ADD COLUMN sentence_date DATE,
+    ADD COLUMN licence_expiry_date DATE,
+    ADD COLUMN sentence_expiry_date DATE,
+    ADD COLUMN sentencing_court VARCHAR(64),
+    ADD COLUMN index_offence VARCHAR(64),
```
jonnywyatt, line 5: From some of the offence descriptions I've seen I think you might need more than 64 characters, but let's find out

mikedunfordEE (approved):

PR #89, opened 2021-08-19
PUD-390: Add booking number

philsegal:


src/main/resources/db/migration/V16__Add_recall_booking_number_column.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE recall ADD COLUMN booking_number VARCHAR(32);
```
mr-ken, line 1: Why not use `TEXT` type (is 32 byte limit meaningful?)

philsegal: We are using VARCHAR in all other places except where we expect free form text. Once we know what the booking number looks like we can restrict the size further if needed

mr-ken (approved):

PR #91, opened 2021-08-20
PUD-434: Fix datetime format to be in UTC.

philsegal:
- Set locale for app as UTC
- Use OffsetDateTime rather than ZonedDateTime

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/config/LocaleConfig.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/config/LocaleConfig.kt
@@ -1,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.managerecallsapi.config
+
+import org.springframework.context.annotation.Configuration
+import java.time.ZoneOffset
+import java.util.TimeZone
+import javax.annotation.PostConstruct
+
+@Configuration
+class LocaleConfig {
+  @PostConstruct
+  fun init() {
+    TimeZone.setDefault(TimeZone.getTimeZone(ZoneOffset.UTC))
+  }
+}
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/integration/RecallsIntegrationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/integration/RecallsIntegrationTest.kt
@@ -4,6 +4,7 @@ import com.natpryce.hamkrest.assertion.assertThat
 import com.natpryce.hamkrest.equalTo
 import com.ninjasquad.springmockk.MockkBean
 import io.mockk.every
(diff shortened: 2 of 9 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/integration/RecallsIntegrationTest.kt
```diff
@@ -132,7 +133,7 @@ class RecallsIntegrationTest : IntegrationTestBase() {
       .jsonPath("$.recallLength").isEqualTo("FOURTEEN_DAYS")
       .jsonPath("$.agreeWithRecallRecommendation").isEqualTo(agree)
       .jsonPath("$.lastReleasePrison").isEqualTo(prison)
-      .jsonPath("$.recallEmailReceivedDateTime").isNotEmpty
+      .jsonPath("$.recallEmailReceivedDateTime").value(endsWith("Z"))
```
AndyNewton, line 136: Do we have a test for this in this build for anything more specific than `endsWith("Z")`?
I'd like us to have one, though I know that can get a bit circular.

mikedunfordEE, line 136: Would be nice if we could assert on actual values here (using Fixed clock or joda time), and possibly see the output of the response json so we can see when/if the format changes.  But possibly one for after our testing discussion next week

mikedunfordEE (approved):
Nice.  Would be nice if we could assert on actual date time values in the test th

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/config/LocaleConfig.kt
```diff
@@ -0,0 +1,16 @@
...
+import java.util.Date
+import java.util.TimeZone
+import javax.annotation.PostConstruct
+
+@Configuration
+class LocaleConfig {
+  @PostConstruct
+  fun init() {
+    TimeZone.setDefault(TimeZone.getTimeZone(ZoneOffset.UTC))
+    System.out.println("Date in UTC: " + Date().toString())
```
mr-ken, line 14: Please remove sysout

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/integration/RecallsIntegrationTest.kt
philsegal: AFAIAA, this is the only place in the code where we validate the json. This uses the random data generation. It is possible we should get rid of that, but I didn't want to do that as part of this ticket. I did read the PACT docs that also said you should avoid using random data, so it could be that we fix the test data and then we can assert on it. Happy to raise a ticket. @mikedunfordEE you raised the same thing

mr-ken (approved):

PR #96, opened 2021-08-24
DT-2297 - randomise timing of circleci security workflow due to rate limiting at veracode.

mattops:


.circleci/config.yml
```diff
@@ -188,7 +188,7 @@ workflows:
   security:
     triggers:
       - schedule:
-          cron: "0 7 * * 1-5"
+          cron: "3 5 * * 1-5"
```
philsegal, line 191: If we want it to be 'random', should we use H?

petergphillips (approved):
LGTM

PR #100, opened 2021-08-24
PUD-330 - Use a real database in `RecallRepositoryIntegrationTest`

dazoakley:


.tool-versions
```diff
@@ -0,0 +1 @@
+java adoptopenjdk-16.0.2+7
```
dazoakley, line 1: This came from Andy's notes on how he set up his macbook: https://mojdt.slack.com/archives/C025S4PG5FT/p1627481715034000?thread_ts=1627383933.012900&cid=C025S4PG5FT

mr-ken (commented):
I think there's a better way than that which doesn't require us to run docker-compose in the background.

dazoakley:
Closing this in favour of Ken's (better) solution. :)

dazoakley:
Re-opened as dynamically spinning up a postgres container in CircleCI is problematic (https://github.com/testcontainers/testcontainers-java/issues/1014).

>Effectively the problem is that the containers initiated inside CircleCI are not accessible from the main container.
If any one is interested in this, it would be good to vote for this to be implemented directly from CircleCI.

(ref https://github.com/testcontainers/testcontainers-java/issues/1014#issuecomment-492872176)

mr-ken:
The other solution requires to run the build with a machine executor (rather than docker) - which is a bit slower and more expensive so although it's a less fussy solution (doesn't require docker-compose) this solution is the more pragmatic one.

mr-ken (approved):

PR #118, opened 2021-08-31
PUD-342 Ability to update "letter to prison" recall fields

mr-ken:


src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/TestFixtures.kt
```diff
@@ -71,7 +71,11 @@ fun recallWithPopulatedFields(
   reasonsForRecallOtherDetail = randomString(),
   agreeWithRecall = AgreeWithRecall.values().random(),
   agreeWithRecallDetail = randomString(),
-  currentPrison = randomString()
+  currentPrison = randomString(),
+  additionalLicenceConditions = true,
```
AndyNewton, line 75: (I thought we had boolean properties before this one but ...).  The pattern in this test for random valid data to be used.  Should we not stick with that pattern in these new properties?

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/TestFixtures.kt
```diff
@@ -71,7 +71,11 @@ fun recallWithPopulatedFields(
...
+  additionalLicenceConditionsDetail = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.",
+  differentNomsNumber = true,
+  differentNomsNumberDetail = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua."
```
AndyNewton, line 78: A good example for me of the value of random data: we're using the same string to test this and `additionalLicenceConditionsDetail ` - with the implication of course that the code might confuse them but the test could still pass.

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/UpdateRecallComponentTest.kt
```diff
@@ -178,4 +184,53 @@ class UpdateRecallComponentTest : ComponentTestBase() {
...
+  private fun invalidLetterToPrisonFields(): Stream<String> {
+    return Stream.of(
+      """{"additionalLicenceConditions": "yes"}""",
+      """{"differentNomsNumber": "yes"}""",
+    )
+  }
+
+  @ParameterizedTest
+  @MethodSource("invalidLetterToPrisonFields")
+  fun `validates the 'letter to prison' boolean fields`(requestJson: String) {
```
AndyNewton, line 223: This looks like a test which can become completely generic?  Maybe a //TODO suggestion?

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/integration/db/RecallRepositoryIntegrationTest.kt
```diff
@@ -79,7 +79,11 @@ class RecallRepositoryIntegrationTest(
       licenceConditionsBreached = "Breached by blah blah blah",
       reasonsForRecall = setOf(ReasonForRecall.ELM_FURTHER_OFFENCE),
       reasonsForRecallOtherDetail = "Because of something else",
-      currentPrison = "MWI"
+      currentPrison = "MWI",
+      additionalLicenceConditions = true,
+      additionalLicenceConditionsDetail = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.",
+      differentNomsNumber = false,
+      differentNomsNumberDetail = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua."
```
AndyNewton, line 86: As above I'm not a fan of using the same test value for two strings.

AndyNewton (approved):
Approved though I'd like to see some tweaks to the tests, as per comments.

PR #124, opened 2021-09-01
PUD-461:  Start postgres when running pact tests

mikedunfordEE:


philsegal (approved):

mr-ken (commented):
Aren't we using MockkBean in the pact test? Why do we need a Postgres instance?

PR #127, opened 2021-09-02
PUD-461:  add missing component tests for recalls not found.  Add error…

mikedunfordEE:
… handling for ResponseStatusExceptions.  remove ManageRecallsApi interface as it wasn't much use

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/db/RecallRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/db/RecallRepository.kt
@@ -29,6 +29,6 @@ class RecallRepository(
     try {
       getById(recallId.value)
     } catch (e: JpaObjectRetrievalFailureException) {
-      throw RecallNotFoundException("Recall not found: '$recallId'", e)
+      throw RecallNotFoundException(recallId)
     }
 }
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallsControllerTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallsControllerTest.kt
@@ -19,7 +19,6 @@ import uk.gov.justice.digital.hmpps.managerecallsapi.db.RecallRepository
 import uk.gov.justice.digital.hmpps.managerecallsapi.domain.NomsNumber
 import uk.gov.justice.digital.hmpps.managerecallsapi.domain.RecallId
 import uk.gov.justice.digital.hmpps.managerecallsapi.domain.random
-import uk.gov.justice.digital.hmpps.managerecallsapi.service.RecallDocumentNotFoundException
 import uk.gov.justice.digital.hmpps.managerecallsapi.service.RecallDocumentService
 import uk.gov.justice.digital.hmpps.managerecallsapi.service.RecallNotFoundException
 import uk.gov.justice.digital.hmpps.managerecallsapi.service.RevocationOrderService
@@ -151,8 +150,7 @@ class RecallsControllerTest {
     val document = "a document"
     val documentBytes = document.toByteArray()
     val category = RecallDocumentCategory.PART_A_RECALL_REPORT
-    val cause = Throwable()
-    val recallNotFoundError = RecallNotFoundException("boom!", cause)
+    val recallNotFoundError = RecallNotFoundException(recallId)
 
     every { recallDocumentService.addDocumentToRecall(recallId, documentBytes, category) } throws recallNotFoundError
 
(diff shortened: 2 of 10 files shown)


philsegal (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/db/RecallRepository.kt
```diff
@@ -29,6 +29,6 @@ class RecallRepository(
     try {
       getById(recallId.value)
     } catch (e: JpaObjectRetrievalFailureException) {
-      throw RecallNotFoundException("Recall not found: '$recallId'", e)
```
mr-ken, line 32: Don't see any issue about being explicit with an error message.

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallsControllerTest.kt
```diff
@@ -168,38 +166,6 @@ class RecallsControllerTest {
...
-
-    val exception = assertThrows<ResponseStatusException> {
-      underTest.getRecallDocument(::RecallId.random(), UUID.randomUUID())
-    }
-
-    assertThat(exception.status, equalTo(HttpStatus.NOT_FOUND))
-    assertThat(exception.reason, equalTo(recallDocumentNotFoundError.message))
-    assertThat(exception.cause, equalTo(recallDocumentNotFoundError))
-  }
-
```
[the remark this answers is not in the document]
mikedunfordEE: The getRecallDocument method doesn't throw a ResponseStatusException anymore, the RecallDocumentService throws the RecallDocumentNotFoundException which is then caught by the ExceptionHandler and tested in the ExceptionHandler test

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/db/RecallRepository.kt
mikedunfordEE: Seems a lot simpler to just contain the relevant ID of the object not found.  When we log it the type of the exception is logged with the recallId which explains clearly what has not been found.  e.g.
`2021-09-02 15:39:33.979  INFO 47841 --- [o-auto-1-exec-8] d.h.m.c.ManageRecallsApiExceptionHandler : RecallNotFoundException(recallId=163abd37-11d7-4954-8de2-ee8ed9c7c98f)`

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallsControllerTest.kt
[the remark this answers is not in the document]
mr-ken: Fair enough - I couldn't find a test in `ExceptionHandlerTest` for `RecallDocumentNotFoundException` though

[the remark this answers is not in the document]
mikedunfordEE: The test is for the base class NotFoundException, could easily be extended to be a parameterised test for all subclasses, depends how rigorous we want to be

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/db/RecallRepository.kt
mikedunfordEE: To me a type that clearly explains what its purpose is and to what it relates is much simpler, and much easier to search through the logs in a consistent manner, a random message string is more prone to be different in different NotFoundException subclasses and hence harder to do more complicated log queries that span more than one exception type (if ever needed).

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallsControllerTest.kt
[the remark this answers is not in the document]
mr-ken: I see - I didn't expect `RecallDocumentNotFoundException` to extend `NotFoundException`

mr-ken (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/db/RecallRepository.kt
mr-ken: I guess it depends how you look at the logs ; when I use Kibana for example I will generally just look at the log level and the error message. This format forces you to also display the stack trace to see the type of the exception, doesn't it?

mikedunfordEE: The error message in this case would be: 
**RecallNotFoundException(recallId=163abd37-11d7-4954-8de2-ee8ed9c7c98f)** 
Just the toString of the RecallNotFoundException data class.  It would be better if we were logging them as json, more structured logging makes complicated querying/analysis of logs a lot easier

PR #128, opened 2021-09-02
PUD-439: Add recallNotificationEmailSentDateTime 

NataliaOsk:


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/UpdateRecallController.kt
```diff
@@ -62,7 +62,8 @@ data class UpdateRecallRequest(
   val additionalLicenceConditions: Boolean? = null,
   val additionalLicenceConditionsDetail: String? = null,
   val differentNomsNumber: Boolean? = null,
-  val differentNomsNumberDetail: String? = null
+  val differentNomsNumberDetail: String? = null,
+  val recallNotificationEmailSentDateTime: LocalDate? = null
```
mr-ken, line 66: Is this a date or date+time. If you need the date and the time in UTC, use `Instant` instead of `LocalDate`

mr-ken (commented):
I'm confused whether

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/UpdateRecallController.kt
philsegal: Actually, we are using OffsetDateTime everywhere else inc `recallEmailReceivedDateTime` in here. But agree it shouldn't be LocalDate as that is just a date, without the time.

philsegal:
There should be an update to `GetRecallComponentTest` as well, to test that the new value is being written as Json. Use `.value(endsWith("Z"))` at least, to 'validate' that we are writing the date time in UTC

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/UpdateRecallController.kt
NataliaOsk: Ye, true. I changed it in the last commits. Just realised quite a lot of commits were not pushed ... silly

philsegal (approved):

mr-ken (approved):
👍🏻

PR #132, opened 2021-09-03
PUD-153: Add status to allow viewing which recalls need dossier created

philsegal:


mr-ken (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallsController.kt
```diff
@@ -180,7 +180,11 @@ data class RecallResponse(
   val differentNomsNumber: Boolean? = null,
   val differentNomsNumberDetail: String? = null,
   val recallNotificationEmailSentDateTime: OffsetDateTime? = null
-)
+) {
+  val status: Status? = calculateStatus()
+
+  private fun calculateStatus() = if (recallNotificationEmailSentDateTime == null) null else Status.RECALL_NOTIFICATION_ISSUED
```
mikedunfordEE, line 186: more kotliny would be:
`recallNotificationEmailSentDateTime?.let { Status.RECALL_NOTIFICATION_ISSUED }`

And if I were being picky you could set status in the toResponse() function to keep the data class a bit cleaner.  A unit test would be nice as well ;-)  

I'll approve anyway, fix forwards if you feel the need!

mikedunfordEE (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallsController.kt
philsegal: ha, I was just adding a unit test. Will take your comment about the`kotliny` improvement. I disagree on the `toResponse` as we'd want all Responses to have a status even if someone used a different method to create one

mikedunfordEE (dismissed):

mikedunfordEE (approved):

PR #135, opened 2021-09-06
PUD-508  Add get dossier end-point on BE to return a PDF

AndyNewton:
PUD-508  Add get dossier end-point on BE to return a PDF - just the RecallNotification as a place-holder for now


AndyNewton:
Would be good to verify PACT versus client PR/branch before merging.
Real work of returning 3 docs pushed to PUD-521.

PR #140, opened 2021-09-08
PUD-521  Introduce documentService getDocumentWithCategory

AndyNewton:
Introduce documentService getDocumentWithCategory; modify some existing tests for class

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallDocumentService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallDocumentService.kt
@@ -49,8 +49,17 @@ class RecallDocumentService(
     val bytes = s3Service.downloadFile(documentId)
     return Pair(document, bytes)
   }
+
+  fun getDocumentWithCategory(recallId: RecallId, documentCategory: RecallDocumentCategory): Pair<RecallDocument, ByteArray> {
+    // For any occurrence of > 1 doc matching recallId and category the actual returned doc here is undefined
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallDocumentService.kt
```diff
@@ -49,8 +49,17 @@ class RecallDocumentService(
     val bytes = s3Service.downloadFile(documentId)
     return Pair(document, bytes)
   }
+
+  fun getDocumentWithCategory(recallId: RecallId, documentCategory: RecallDocumentCategory): Pair<RecallDocument, ByteArray> {
+    // For any occurrence of > 1 doc matching recallId and category the actual returned doc here is undefined
```
mr-ken, line 54: Isn't there a unique constraint in the recall_document table for recall_id/category?

mr-ken (approved):
Makes sense

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallDocumentService.kt
AndyNewton: You're right @mr-ken ; there is; I didn't spot it before: https://github.com/ministryofjustice/manage-recalls-api/blob/main/src/main/resources/db/migration/V4__Create_recall_document_table.sql#L14
That's why I was asking on Slack earlier.
I'll edit my comment.

PR #142, opened 2021-09-09
Pud 517 update document

mikedunfordEE:


.circleci/config.yml
```diff
@@ -34,12 +42,19 @@ executors:
...
+      - hmpps/install_aws_cli
+      - hmpps/wait_till_ready:
+          port: 4566
+      - run:
+          name: Create S3 bucket
+          command: |
+            export AWS_ACCESS_KEY_ID=dontcare
+            export AWS_SECRET_ACCESS_KEY=dontcare
+            export AWS_DEFAULT_REGION=eu-west-2
+            aws s3api create-bucket --bucket test-manage-recalls-api --region eu-west-2 --endpoint-url http://localhost:4566
```
philsegal, line 57: minor, but not sure you need region here if you have the env var

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/db/RecallRepository.kt
```diff
@@ -31,4 +34,17 @@ class RecallRepository(
     } catch (e: JpaObjectRetrievalFailureException) {
       throw RecallNotFoundException(recallId)
     }
+
+  override fun findByRecallId(recallId: RecallId): Recall? =
+    findById(recallId.value).orElse(null)
+
+  @Transactional
+  override fun addDocumentToRecall(recallId: RecallId, recallDocument: RecallDocument) {
+    getByRecallId(recallId).let { recall ->
```
philsegal, line 43: what happens if we dont find a recall?

philsegal (approved):

.circleci/config.yml
mikedunfordEE: ok will try without

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/db/RecallRepository.kt
mikedunfordEE: getRecallById throws a RecallNotFoundException, although it is guarded from that happening in the RecallDocumentService.  Could add an extra unit test to document that though

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/db/RecallRepository.kt
```diff
@@ -12,13 +12,16 @@ import org.springframework.stereotype.Repository
...
 import java.util.UUID
+import javax.transaction.Transactional
 
 @Repository("jpaRecallRepository")
 interface JpaRecallRepository : JpaRepository<Recall, UUID>
 
 @NoRepositoryBean
 interface ExtendedRecallRepository : JpaRecallRepository {
   fun getByRecallId(recallId: RecallId): Recall
+  fun findByRecallId(recallId: RecallId): Recall?
```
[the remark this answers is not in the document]
mikedunfordEE: The 2 methods intentionally behave differently.  'getRecallById' throws a RecallNotFoundException if a recall doesn't exist.  'findByRecallId' returns null if it doesn't find one.   It's standard practice to have a get and a find that behave this way.  e.g. the JpaRecallRepository interface has a getById(UUID) and findById(UUID) that behave the same way.  The methods I have added are just specific to be used with a RecallId instead of UUID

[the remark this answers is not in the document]
mr-ken: fair enough - thanks for clarifying

mr-ken (approved):
Looks good

PR #144, opened 2021-09-10
Cleanup

philsegal:
PUD-488: Move revocation-order-logo.png
PUD-XXX: Ensure that any new field of UpdateRecallRequest is tested by constructing a 'full' recall using reflection.
- tidy up duplication in UpdateRecallControllerTest


mr-ken (commented):
This feels like a trade-off between simple/boiler plate <-> complex
If we want to go the property-based testing way there some good solutions in Kotlin like Kotest (https://kotest.io/docs/proptest/property-based-testing.html) for that.

mikedunfordEE (approved):

PR #156, opened 2021-09-15
PUD-438: Build the request metrics dashboards into the helm chart

dazoakley:
This is a temporary commit to prove out the work I've been doing on the `generic-service` helm chart.

Basically it's a copy/paste of the `generic-service` chart and I've added in the following:

- An optional [traefik](https://traefik.io/) proxy in front of the main container (for "internal" services to get request metrics)
- A golden signals dashboard using the metrics coming from Traefik
- A golden signals dashboard using the metrics coming from the NGINX ingress

We don't really need two (almost identical) dashboards, they are both here for testing purposes. In the final cut we would only have one based on if the traefik proxy is needed or not (if we have an ingress, it's not).


mr-ken (approved):
That looks really cool @dazoakley 
So is the idea to incorporate this into the Kotlin microservice template?

dazoakley:
Cheers :)

Nope, this is an addition to the `generic-service` helm chart (that both the kotlin and typescript template projects use), so basically all users of the templates and helm charts can get standardised dashboarding.

PR #161, opened 2021-09-15
PUD-164: Add page numbering. Validate number of pages generated by gotenberg

philsegal:
Currently we are not validating that the numbers have been added since the bytes seem to vary with each execution and so cant be verified. Potentially we can look for an alternative way to validate.


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/documents/PdfDecorator.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.lowagie.text.pdf.PdfReader
+import com.lowagie.text.pdf.PdfStamper
+import org.librepdf.openpdf.fonts.Liberation
+import org.springframework.stereotype.Component
+import java.io.ByteArrayOutputStream
+
+@Component
+class PdfDecorator {
+
+  fun numberPages(pdfContent: ByteArray): ByteArray {
```
AndyNewton, line 15: Given we have no unit test of this code how about the ugly solution of a comment like e.g.:
  // Note: this function is tested within CreateDossierComponentTest

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/DossierServiceTest.kt
```diff
@@ -20,8 +21,9 @@ internal class DossierServiceTest {
   private val revocationOrderService = mockk<RevocationOrderService>()
   private val pdfDocumentGenerator = mockk<PdfDocumentGenerator>()
   private val recallDocumentService = mockk<RecallDocumentService>()
+  private val pdfDecorator = mockk<PdfDecorator>()
 
-  private val underTest = DossierService(revocationOrderService, pdfDocumentGenerator, recallDocumentService)
+  private val underTest = DossierService(revocationOrderService, pdfDocumentGenerator, recallDocumentService, pdfDecorator)
 
   @Test
   fun `get dossier returns part A, license and revocation order as dossier when all present for recall`() {
```
AndyNewton, line 29: You'll love this: add "as numbered pages" to test title?

AndyNewton (approved):
Approved, but I wouldn't mind a couple of comments added - given the absence of a unit test for the numberer.

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/documents/PdfDecorator.kt
philsegal: But its not tested :) its mocked in that test

mr-ken (approved):
Looks good

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/DossierServiceTest.kt
philsegal: we're not numbering it tho, its mocked.

AndyNewton: Yeah, but the same is true of every step in this test.  IMO, it's a pretty weak test.  And yes, I wrote the original version.  The component test comes to the rescue fortunately.

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/documents/PdfDecorator.kt
AndyNewton: Where's :mr-hankey: when you need him?

AndyNewton: OK.  I've added a unit test for `fun numberPages` now: https://github.com/ministryofjustice/manage-recalls-api/pull/165

PR #164, opened 2021-09-15
Set the SENTRY_RELEASE env variable

dazoakley:
This will allow us to match up errors to reported releases.

ref PUD-601


AndyNewton (approved):
LGTM!

PR #175, opened 2021-09-20
PUD-540:  Add assessedByUserId field to Recall to satisfy the pact test

mikedunfordEE:


src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/pact/provider/ManageRecallsUIPactTest.kt
```diff
@@ -191,7 +191,7 @@ class ManagerRecallsUiUnauthorizedPactTest : ManagerRecallsUiPactTestBase() {
 @VerificationReports(value = ["console"])
 @Provider("manage-recalls-api")
 @Consumer("manage-recalls-ui")
-@PactBroker
+@PactFolder("../manage-recalls-ui/pact/pacts")
```
AndyNewton, line 194: Would be good to have a cleaner switch than this of course!

AndyNewton (changes requested):
This has sneaked in ... `@PactFolder("../manage-recalls-ui/pact/pacts")`

AndyNewton (approved):

PR #178, opened 2021-09-21
Move the golden signals dashboard

dazoakley:
After chatting with the folks from the DPS team we're going to try putting the golden signals dashboard in with the prometheus alerts.  

This stops the main `generic-service` chart from getting a bit bloated, but it should also let us match the alert thresholds in the actual alerts up with thresholds in the grafana charts (in a future commit).


mr-ken (approved):
Looks good @dazoakley 
Maybe a silly question but where are we moving the dashboards from?

dazoakley:
👍  

I'd originally added them to the `generic-service` chart but there was some resistance to adding extra code to that.

mr-ken:
gotcha - thanks

PR #181, opened 2021-09-21
Make the dashboard UID generation a bit more complicated... :(

dazoakley:
`manage-recalls-api` and `manage-recalls-ui` were producing the same UID
with the old logic (as they were essentially the same string up until
the last 2-3 characters - and this got truncated).


helm_deploy/manage-recalls-api/values.yaml
```diff
@@ -39,5 +39,5 @@ generic-service:
       S3_BUCKET_NAME: "bucket_name"
 
 generic-prometheus-alerts:
-  targetApplication: manage-recalls-api
+  targetApplication: manage-recalls-api-foo
```
AndyNewton, line 42: Intentional `-foo` or unintentional?

dazoakley: 😆 good spot!

AndyNewton (approved):

PR #184, opened 2021-09-22
PUD-541:  Add dossierCreatedByUserId to Recall

mikedunfordEE:


mr-ken (commented):
The purpose of using an Auth token is to identify/authenticate the user making the request so passing a user ID doesn't make much sense to me ; especially since you can pass anything you want (there's no validation). I can see all kinds of issues popping up because of this...

dazoakley (approved):

mikedunfordEE:
> The purpose of using an Auth token is to identify/authenticate the user making the request so passing a user ID doesn't make much sense to me ; especially since you can pass anything you want (there's no validation). I can see all kinds of issues popping up because of this...

Yes 100% agree.  However at the moment the auth token does not contain the UUID we want to use, so we decided for now to just pass through from the UI as it already has access to the UUID to keep things simple.   Definitely not the long term solution, hopefully the dps_tech_team will add the UUID to the token soon (ticket has been raised) and we can review these stories, and also hopefully put some proper auditing of user actions in place

mr-ken:
> > The purpose of using an Auth token is to identify/authenticate the user making the request so passing a user ID doesn't make much sense to me ; especially since you can pass anything you want (there's no validation). I can see all kinds of issues popping up because of this...
> 
> Yes 100% agree. However at the moment the auth token does not contain the UUID we want to use, so we decided for now to just pass through from the UI as it already has access to the UUID to keep things simple. Definitely not the long term solution, hopefully the dps_tech_team will add the UUID to the token soon (ticket has been raised) and we can review these stories, and also hopefully put some proper auditing of user actions in place

I think it would have been simpler and safer to make a request to the Auth service to get the user ID rather than introduce new fields which means changing the contract + passing on changes the clients. Also it's introducing a dependency on another team (which we said we wouldn't do) ; we aren't even sure how much effort it is to include the user ID to the access token.

PR #186, opened 2021-09-22
PUD-657: Update the flush script

dazoakley:
* Don't error out if the port-forward is already present
* Check for system dependencies
* Also clear down the S3 bucket


scripts/flush-dev-db.sh
```diff
@@ -1,18 +1,49 @@
...
+# Check dependencies
+check_dep "psql" "brew install postgresql"
+check_dep "jq" "brew install jq"
+check_dep "aws" "brew install awscli"
+check_dep "kubectl" "asdf install kubectl 1.19.15"
+
+# check for the ENV variable
+set +u
+if [[ -z "${ENV}" ]]; then
+  echo "You must set the ENV variable - allowed values: [dev, preprod, prod]"
```
philsegal, line 20: prod seems pretty gnarly/dangerous. surely we cant/shouldnt have access to prod?

dazoakley: Fair point - will restrict to dev or pre-prod 👍

scripts/flush-dev-db.sh
```diff
@@ -1,18 +1,51 @@
...
+
+# Check dependencies
+check_dep "psql" "brew install postgresql"
+check_dep "jq" "brew install jq"
+check_dep "aws" "brew install awscli"
+check_dep "kubectl" "asdf install kubectl 1.19.15"
+
+# check for the ENV variable
+set +u
+if [[ ! "${ENV}" =~ ^(dev|preprod)$ ]]; then
```
AndyNewton, line 19: Personally I'd prefer this and `port-forward-db` to both use a CL arg for env not an environment variable - but also they should be consistent with one another.

scripts/flush-dev-db.sh
```diff
@@ -1,18 +1,51 @@
 #!/bin/bash
```
AndyNewton, line 1: For me the name of this script is a little misleading: I could do without 'dev' in the name: suggest to me that it acts on the dev DB only.

AndyNewton (commented):
Couple of comments for discussion.

philsegal (approved):

PR #189, opened 2021-09-23
PUD-468: Add signature to user details and use in revocation order.

philsegal:


mikedunfordEE (approved):

mr-ken (commented):
I think this code change would have done well with another pair of eyes tbh...
- why does the consumer now need to know know a user ID (whose user ID?) in order to download a revocation order? The user ID should be persisted in the database. Why not get it from there? Unless I'm missing something...
- why not store the signature, an image file, like any other file in S3? This will bloat the database table unnecessarily and it's not much more effort. It also might be a requirement to scan that file for security vulnerabilities

PR #190, opened 2021-09-23
PUD-644:  Rename / restructure classes used to send data to gotenberg

mikedunfordEE:


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/documents/PdfDocumentGenerationService.kt
```diff
@@ -59,24 +65,36 @@ class PdfDocumentGenerationService(@Autowired private val gotenbergApi: Gotenber
...
 }
 
-data class ClassPathDocumentDetail(override val name: String, val path: String = "/templates/images/$name") :
-  DocumentDetail<MultipartInputStreamFileResource> {
-  override fun data() = MultipartInputStreamFileResource(ClassPathResource(path).inputStream, name)
+interface ImageData<T> : Data<T> {
+  val name: String
+}
+
+data class ClassPathImageData(override val name: String) :
```
AndyNewton, line 76: I'm going to stop worrying about it but this `name` property is actually a `filename`; whereas the `filename` property of `class MultipartInputStreamFileResource` is ...?  Right now I can't see why we still need that.  Must be time for me tea.

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/matchers/MatchersHamkrest.kt
```diff
@@ -39,3 +41,22 @@ fun isPdfWithNumberOfPages(numberOfPagesMatcher: Matcher<Int>): Matcher<ByteArra
         numberOfPagesMatcher(pdfReader.numberOfPages)
       }
   }
+
+fun <T> onlyContainsInOrder(expected: List<Matcher<T>>): Matcher<List<T>> =
```
AndyNewton, line 45: Seems bizarre that we have to implement this for ourselves.

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallNotificationServiceTest.kt
```diff
@@ -36,32 +41,46 @@ internal class RecallNotificationServiceTest {
...
+        listOf(
+          documentWith(recallSummaryContent),
+          documentWith(revocationOrderContent),
+          documentWith(letterToProbationContent)
+        )
+      )
+    )
   }
 
+  private fun documentWith(documentContent: String) =
```
AndyNewton, line 74: Two copies of this: just wondering what the "Kotlinic"? way to share them might be?

AndyNewton (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/documents/PdfDocumentGenerationService.kt
```diff
@@ -59,24 +65,36 @@ class PdfDocumentGenerationService(@Autowired private val gotenbergApi: Gotenber
...
 
-data class InputStreamDocumentDetail(override val name: String, val inputStream: InputStream) :
-  DocumentDetail<MultipartInputStreamFileResource> {
-  override fun data() = MultipartInputStreamFileResource(inputStream, name)
+data class InputStreamDocumentData(val inputStream: InputStream) : Data<MultipartInputStreamFileResource> {
+  override fun data() = MultipartInputStreamFileResource(inputStream)
 }
 
-class MultipartInputStreamFileResource(inputStream: InputStream, private val filename: String) :
+class MultipartInputStreamFileResource(inputStream: InputStream, private val filename: String? = null) :
```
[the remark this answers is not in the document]
mikedunfordEE: yeah could be

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/documents/PdfDocumentGenerationService.kt
mikedunfordEE: it's also the name of the image in the HTML, so needs to match the HTML.  Might do as you suggested yesterday and try and and some sort of type to encapsulate the Image name so we can reuse in the HTML generation and the PDF generation, so this might disappear

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/matchers/MatchersHamkrest.kt
mikedunfordEE: Yes, there are ways to almost do it, but not quite, see this thread:  https://github.com/npryce/hamkrest/issues/37

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallNotificationServiceTest.kt
mikedunfordEE: Yes I was lazy here, thought about reusing it but went for the cut and paste option.  I'll move somewhere when I next come across it

PR #209, opened 2021-09-28
Pud 151 Add recall notification page count to recall summary

mikedunfordEE:


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallSummaryService.kt
```diff
@@ -14,6 +15,9 @@ import uk.gov.justice.digital.hmpps.managerecallsapi.search.Prisoner
 import uk.gov.justice.digital.hmpps.managerecallsapi.search.PrisonerOffenderSearchClient
 import uk.gov.justice.digital.hmpps.managerecallsapi.service.RecallImage.HmppsLogo
 
+// This is the number of pages of the Revocation Order and Letter to Probation
+const val OTHER_PAGES_IN_RECALL_NOTIFICATION = 2
```
philsegal, line 19: can we be certain that these pages are always a single page?

mikedunfordEE: Yes, I confirmed with Ollie

philsegal (approved):

PR #218, opened 2021-09-30
Pud 62 Add details to Letter to Probation

mikedunfordEE:


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallSummaryGenerator.kt
```diff
@@ -16,15 +16,15 @@ import java.time.LocalDate
...
+import java.util.Locale.ENGLISH
 
 @Component
 class RecallSummaryGenerator(
   @Autowired private val templateEngine: SpringTemplateEngine,
   @Autowired private val clock: Clock
 ) {
 
-  private val dateFormatter = DateTimeFormatter.ofPattern("dd MMM yyyy", Locale.ENGLISH)
+  private val dateFormatter = DateTimeFormatter.ofPattern("dd MMM yyyy", ENGLISH)
```
AndyNewton, line 27: Can supersede with your new `.asStandardDate`?
Also, interesting use of a locale here and not elsewhere??

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallSummaryGenerator.kt
```diff
@@ -16,15 +16,15 @@ import java.time.LocalDate
...
   private val timeFormatter = DateTimeFormatter.ofPattern("HH:mm")
```
AndyNewton, line 28: Ditto `asStandardTime`?

AndyNewton (approved):
Approved with two minor QQs - can be for following PR.

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallSummaryGenerator.kt
mikedunfordEE: Yes I intend to reuse all over the place.  Just didn't want to change too much while others were still working in this area.

Without being specific about the Locale I was getting differences locally and in circleCI e.g. Sep vs Sept

PR #219, opened 2021-09-30
PUD-687: Populate Local Delivery Unit in recall summary

philsegal:
- add reference data controller with LDU endpoint


src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/AuthenticatedClient.kt
```diff
@@ -103,6 +104,12 @@ class AuthenticatedClient(
   fun getUserDetails(userId: UserId) =
     getRequest("/users/${userId.value}", UserDetailsResponse::class.java)
 
+  fun localDeliveryUnit(): List<LocalDeliveryUnitResponse> =
+    sendGetRequest("/reference-data/local-delivery-unit")
```
jonnywyatt, line 108: should be plural? /local-delivery-units

philsegal: I thought that, then thought we were doing singular...1 second...

jonnywyatt: 👍  I won't approve the rest, will leave that to someone who knows Kotlin

mikedunfordEE (approved):

PR #220, opened 2021-09-30
PUD-62:  Include middleNames in PersonName

mikedunfordEE:


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/domain/domain.kt
```diff
@@ -14,6 +14,7 @@ class NomsNumber(value: String) : Validated<String>(value, notBlank, alphanumeri
 class RecallId(value: UUID) : Validated<UUID>(value)
 class UserId(value: UUID) : Validated<UUID>(value)
 class FirstName(value: String) : Validated<String>(value, notBlank)
+class MiddleNames(value: String) : Validated<String>(value, notBlank)
```
philsegal, line 17: Does it need to be plural? 
FirstName can already be multiple names with a space, as can surname?

philsegal (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/domain/domain.kt
mikedunfordEE: Yes, you can have more than one middle name.  Trying to be specific about what the type is

philsegal: I get that, but you can have more than one first name or surname.

PR #232, opened 2021-10-04
Remove the dependency on the trivy scan (temporarily) to deploy

dazoakley:


philsegal (approved):

dazoakley:
No need - trivy is working now

PR #238, opened 2021-10-05
Pud 731 dossier context factory creation

mikedunfordEE:


NataliaOsk (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/DossierContextFactory.kt
```diff
@@ -0,0 +1,70 @@
...
+}
+
+data class DossierContext(
+  val recall: Recall,
+  val prisoner: Prisoner,
+  val currentPrisonName: PrisonName,
+) {
+  fun getReasonsForRecallContext(): ReasonsForRecallContext {
+    return ReasonsForRecallContext(
+      FirstAndMiddleNames(FirstName(prisoner.firstName!!), prisoner.middleNames?.let { MiddleNames(it) }),
```
philsegal, line 39: feels like we should add a function for prisoner -> FullName and/or FirstAndMiddleNames

philsegal (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/DossierContextFactory.kt
mikedunfordEE: yes there is one, looks like we didn't use it ;-)

PR #245, opened 2021-10-05
PUD-743: Use the prisoner first and last name if hasPrevConsMainName is false

philsegal:
- standardise PersonName

Will look to make changes so that the value is passed from front end in the next PR


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/documents/recallnotification/RecallNotificationContextFactory.kt
```diff
@@ -60,8 +59,7 @@ data class RecallNotificationContext(
   fun getRevocationOrderContext(): RevocationOrderContext {
     return RevocationOrderContext(
       recall.recallId(),
-      FirstAndMiddleNames(FirstName(prisoner.firstName!!), prisoner.middleNames?.let { MiddleNames(it) }),
-      LastName(prisoner.lastName!!),
+      PersonName(FirstName(prisoner.firstName!!), prisoner.middleNames?.let { MiddleNames(it) }, LastName(prisoner.lastName!!)),
```
AndyNewton, line 62: If we add a secondary constructor to PersonName:
`
  constructor(firstName: String, middleNames: String? = null, lastName: String) :
    this(FirstName(firstName), middleNames?.let { MiddleNames(it) }, LastName(lastName))
`
then all the (17?) calls like this one simplify to e.g.:
`      PersonName(prisoner.firstName!!, prisoner.middleNames, prisoner.lastName!!),`

AndyNewton (commented):
Suggesting a secondary constructor for `PersonName`.

mikedunfordEE (approved):

PR #254, opened 2021-10-08
PUD-323: Add Court Register and validation

philsegal:
- make reference-data endpoint unauthenticated
- add courts and prisons to ref-data endpoint


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/db/Recall.kt
```diff
@@ -175,6 +177,7 @@ data class SentencingInfo(
   val sentenceLength: SentenceLength,
   val conditionalReleaseDate: LocalDate? = null
 ) {
+  fun sentencingCourt(): CourtId = CourtId(sentencingCourt)
```
mikedunfordEE, line 180: sentencingCourt can be CourtId instead of String, see PrisonId

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/domain/domain.kt
```diff
@@ -22,6 +22,8 @@ class Email(value: String) : Validated<String>(value, notBlank)
 class PhoneNumber(value: String) : Validated<String>(value, notBlank)
 class PrisonName(value: String) : Validated<String>(value, notBlank)
 class PrisonId(value: String) : Validated<String>(value, notBlank, alphanumeric, 2.minLength, 6.maxLength)
+class CourtId(value: String) : Validated<String>(value, notBlank)
```
mikedunfordEE, line 25: Option to add extra validation rules here, just need to check the implementation in the court-register repo.  Although the swagger docs and DB field size were not consistent when I checked the prison-register :-(

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/CourtValidationService.kt
```diff
@@ -0,0 +1,16 @@
...
+import org.springframework.stereotype.Service
+import uk.gov.justice.digital.hmpps.managerecallsapi.domain.CourtId
+import uk.gov.justice.digital.hmpps.managerecallsapi.register.court.CourtRegisterClient
+
+@Service
+class CourtValidationService(@Autowired private val courtRegisterClient: CourtRegisterClient) {
+
+  fun isValid(courtId: CourtId?): Boolean {
+    return courtId?.let {
+      courtRegisterClient.getAllCourts().block()?.any { (it.courtId == courtId) } ?: false
```
mikedunfordEE, line 13: should probably do the same as the prison validation and use the get by id endpoint

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/register/court/CourtRegisterClient.kt
```diff
@@ -0,0 +1,29 @@
...
+  fun getAllCourts(): Mono<List<Court>> {
+    return courtRegisterWebClient
+      .get()
+      .uri("/courts/all")
+      .retrieve()
+      .bodyToMono(object : ParameterizedTypeReference<List<Court>>() {})
+  }
+
+  data class Court(
+    val courtId: CourtId? = null,
```
mikedunfordEE, line 26: These are both required fields in the swagger docs, so don't need to be nullable

PR #274, opened 2021-10-14
PUD-736, PUD-737, PUD-744, PUD-756, PUD-730, PUD-753, PUD-754, PUD-755: Fix styling

philsegal:


src/main/resources/templates/recall-summary.html
```diff
@@ -687,7 +688,14 @@
...
   .c7 {
-    font-size: 10pt
+    font-size: 10pt;
+    width: 100%;
+  }
+
+  pre {
+    font-family: "Arial";
+    font-size: 10pt;
+    white-space: pre-wrap;
```
AndyNewton, line 698: I guess this may well all be superseded ... but I wonder why we need `white-space: pre-wrap` some places and `white-space: pre-line` in others? 🤷‍♂️

src/main/resources/templates/letter-to-prison_custody-office.html
```diff
@@ -267,18 +267,16 @@
   <div th:if="${hasAdditionalLicenceConditions}">
     <p class=MsoNormal style='margin-left:0cm;text-indent:-.1pt'>Additional Licence Conditions requested by Offender Manager - Please ensure these are included on the licence when the offender is re-released:</p>
 
-    <p class=MsoNormal style='margin-left:0cm;text-indent:-.1pt'><span th:text="${additionalLicenceConditionsDetail}"/></p>
+    <p class=MsoNormal style='margin-left:0cm;text-indent:-.1pt'><pre class=MsoNormal style='margin-left:0cm;text-indent:-.1pt;white-space: pre-wrap' th:text="${additionalLicenceConditionsDetail}"></pre></p>
```
AndyNewton, line 270: Do we need to reproduce the class setting from the para on the contained pre?  E.g. in this case that gives us 4 copies of `pre class="MsoNormal" style="white-space: pre-wrap`?

src/main/resources/templates/recall-summary.html
```diff
@@ -1012,14 +1020,15 @@
 <table class="c57">
   <tbody>
   <tr class="c49">
-    <td class="c39" colspan="1" rowspan="1"><p class="c0">
-      <p class="c24"><span class="c1">VULNERABILITY ISSUES: <span th:text="${vulnerabilityDiversityDetail}"></span></p>
+    <td class="c39" colspan="1" rowspan="1"><p class="c0 c24">
+        <span class="c1">VULNERABILITY ISSUES: </span>
+        <pre class="c24" th:text="${vulnerabilityDiversityDetail}"/>
```
AndyNewton, line 1025: Just another example where I wouldn't think that we need to copy `c24` from the outer para onto the pre?

philsegal: A fair amount of this was trial and error. And once I got it working I stopped changing stuff, so there could be some unnecessary markup, but it all needs rewriting at some point. However, `pre` doesnt take any styling from parent elements so you need to add all the classes explicitly.

src/main/resources/templates/letter-to-prison_custody-office.html
philsegal: It appeared we do (pre doesnt inherit styling)

PR #281, opened 2021-10-20
PUD-738: Add endpoints for assignee

philsegal:


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallsController.kt
```diff
@@ -92,6 +96,25 @@ class RecallsController(
...
+  fun assignRecall(
+    @PathVariable("recallId") recallId: RecallId,
+    @RequestBody assignRecallRequest: AssignRecallRequest
+  ): RecallResponse =
+    recallRepository.getByRecallId(recallId)
+      .copy(assignee = assignRecallRequest.assignee.value)
+      .let { recallRepository.save(it) }
+      .toResponse()
+
+  @DeleteMapping("/recalls/{recallId}/assignee")
```
AndyNewton, line 110: Think we agreed (just this morning) to require `{assigneeId}` as ending path param.

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallsController.kt
```diff
@@ -92,6 +96,25 @@ class RecallsController(
       )
     )
   }
+
+  @PostMapping("/recalls/{recallId}/assignee")
```
AndyNewton, line 100: Think we agreed (just this morning) to add `{assigneeId}` as ending path param rather than body - for consistency with delete?

AndyNewton (changes requested):

AndyNewton (dismissed):

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/AssignRecallComponentTest.kt
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @Test
+  @Throws(MethodArgumentTypeMismatchException::class)
+  fun `must pass a valid uuid as assignee`() {
+    val recallId = ::RecallId.random()
+
+    authenticatedClient.delete("/recalls/$recallId/assignee/notauuid", INTERNAL_SERVER_ERROR)
+    authenticatedClient.post("/recalls/$recallId/assignee/notauuid", "").expectStatus().isEqualTo(INTERNAL_SERVER_ERROR)
+  }
```
AndyNewton, line 71: Does this really test both operations?

AndyNewton: Approving though would be clearer IMO as two @Tests.

PR #284, opened 2021-10-21
PUD-738: Return assigneeUserName on RecallResponse

philsegal:
- add PACT for assignee


src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/pact/provider/ManageRecallsUIPactTest.kt
```diff
@@ -103,17 +109,31 @@ class ManagerRecallsUiAuthorizedPactTest : ManagerRecallsUiPactTestBase() {
...
 
   @State(
     "a recall exists",
-    "a fully populated recall exists"
+    "a fully populated recall exists",
+    "a user can be unassigned"
   )
   fun `a fully populated recall exists without documents`() {
-    recallRepository.save(fullyPopulatedRecall(::RecallId.zeroes()).copy(documents = emptySet()))
+    val assigneeUuid = UUID.fromString("11111111-1111-1111-1111-111111111111")
```
AndyNewton, line 124: FYI only: we shouldn't have to used a fixed value for this: our PACTs verify presence not values.

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallsControllerTest.kt
```diff
@@ -216,4 +248,56 @@ class RecallsControllerTest {
 
     assertThat(result, equalTo(RecallResponse(recallId, nomsNumber)))
   }
+
+  private val recall = Recall(recallId, nomsNumber)
```
AndyNewton, line 252: I'd rather we didn't paste class variables into the middle of the class but 🤷‍♂️

AndyNewton (approved):
Approved with minor comments.  Will probably merge too.

PR #314, opened 2021-11-03
PUD-816: Enable prometheus metrics and include upstream health metrics

dazoakley:
TODO: Expose the metrics to prometheus once I've confirmed the http request timing is appropriate on the cluster.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/health/BankHolidayHealth.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/health/BankHolidayHealth.kt
@@ -7,5 +7,6 @@ import org.springframework.web.reactive.function.client.WebClient
 @Component("bankHoliday")
 class BankHolidayHealth(
   webClient: WebClient,
+  @Value("bankHoliday") componentName: String,
   @Value("\${bankHolidayRegister.endpoint.url}") endpointUrl: String
-) : PingHealthCheck(webClient, endpointUrl)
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/health/BankHolidayHealth.kt
```diff
@@ -7,5 +7,6 @@ import org.springframework.web.reactive.function.client.WebClient
 @Component("bankHoliday")
 class BankHolidayHealth(
   webClient: WebClient,
+  @Value("bankHoliday") componentName: String,
```
philsegal, line 10: Is there a reason to have the component name dynamic here but static in other places? Be nice to be consistent? Its a shame we cant reuse the `@Component` name

dazoakley: Can you use the name from `@Component` - genuine question as that would be much better... 😄 

This is the first bit o' Kotlin/Java I've done in well over a decade so there's lots about stuff I don't know

philsegal (approved):
Be nice to reuse the ComponentName, but am assuming that wasn't an option.

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/health/BankHolidayHealth.kt
philsegal: Lets merge like this, I'll have a look around but initial query suggests its not an option

PR #332, opened 2021-11-12
PUD-969: Use user_uuid from auth token rather than requiring client to provide it

philsegal:


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallsController.kt
```diff
@@ -48,15 +55,28 @@ class RecallsController(
...
+  fun bookRecall(@RequestBody bookRecallRequest: BookRecallRequest, @RequestHeader("Authorization") bearerToken: String): ResponseEntity<RecallResponse> {
+    val token = getTokenFromHeader(bearerToken)
+
+    return ResponseEntity(
+      recallRepository.save(bookRecallRequest.toRecall(token.userUuid)).toResponse(),
       HttpStatus.CREATED
     )
+  }
+
+  private fun getTokenFromHeader(bearerToken: String): Token {
```
AndyNewton: "Obvs" move this to a helper/the bean/don't copy to each controller - as a future refactoring.

AndyNewton: Also, I guess the "real" test of this is the E2Es.

AndyNewton (approved):
Soz for delay.  Looks OK, albeit a bit 'black magic'.  Couple of minor comments.

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallsController.kt
philsegal: Yep,  have done so in the next commit :)

PR #334, opened 2021-11-12
PUD-956 BE Reduce chance of error in entering index offence

AndyNewton:
PUD-956 BE Reduce chance of error in entering index offence
Ugly but at least a candidate for merge; possible follow-on clean-up.
Includes scripts, README notes and raw input file for reference/next time re. data conversion pain.
Need to verify with PACT and check the result is good for the UI still.


philsegal (changes requested):
Can you add to `EndpointSecurityComponentTest` and add the other missing ref-data ones too ?

AndyNewton:
Addressed `EndpointSecurityComponentTest` request in ^^^ commit @philsegal .

philsegal (approved):

PR #339, opened 2021-11-17
Amend pact to include "wip" pacts. Hoping that this will ensure that …

philsegal:
…when UI PR pact check fails, it will rerun when API is updated.


philsegal:
After discussing with @AndyNewton going to leave this for now

PR #342, opened 2021-11-19
PUD-852: Add MissingDocumentsRecord controller & persistence

philsegal:


src/main/resources/db/migration/V55__Add_missing_documents_record_tables.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE missing_documents_record (
```
AndyNewton, line 1: File name needs to be at least "V56..." now.

src/main/resources/db/migration/V55__Add_missing_documents_record_tables.sql
```diff
@@ -0,0 +1,15 @@
...
+          detail text NOT NULL,
+          version INTEGER NOT NULL,
+          created_by UUID NOT NULL,
+          created_date_time TIMESTAMPTZ NOT NULL
+);
+
+CREATE TABLE missing_document_category (
+    missing_document_record_id    UUID    NOT NULL REFERENCES missing_documents_record (id),
+    document_category   VARCHAR(32) NOT NULL,
+    UNIQUE (missing_document_record_id, document_category)
```
AndyNewton, line 14: I haven't spotted a test verifying this constraint.  For documents we have those in the IntegrationTest.  Tests named like "fun `cannot save and flush ... throws DataIntegrityViolationException`()".

AndyNewton (changes requested):
Two minor QQs.  Let me know also when merge conflicts are resolved please.

src/main/resources/db/migration/V55__Add_missing_documents_record_tables.sql
philsegal: we are saving a set for this. But for other restraints I will add.

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/MissingDocumentsRecordController.kt
```diff
@@ -0,0 +1,91 @@
...
+
+data class MissingDocumentsRecordRequest(
+  val recallId: RecallId,
+  val categories: List<DocumentCategory>,
+  val detail: String,
+  val emailFileContent: String,
+  val emailFileName: String
+)
+
+data class MissingDocumentResponse(val documentId: DocumentId)
```
AndyNewton, line 91: Class unused?

src/main/resources/db/migration/V56__Add_missing_documents_record_tables.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TABLE missing_documents_record (
+          id UUID PRIMARY KEY,
+          recall_id UUID NOT NULL REFERENCES recall (id),
+          email_id UUID NOT NULL REFERENCES document (id),
+          detail text NOT NULL,
```
AndyNewton, line 5: 'TEXT' preferred to 'text' in general - later.

src/main/resources/db/migration/V56__Add_missing_documents_record_tables.sql
```diff
@@ -0,0 +1,16 @@
...
+          version INTEGER NOT NULL,
+          created_by_user_id UUID NOT NULL,
+          created_date_time TIMESTAMPTZ NOT NULL,
+          UNIQUE (recall_id, version)
+);
+
+CREATE TABLE missing_document_category (
+    missing_document_record_id    UUID    NOT NULL REFERENCES missing_documents_record (id),
+    document_category   VARCHAR(32) NOT NULL,
```
AndyNewton, line 14: FYI the other store we have for the same info - I believe - is defined as `category  TEXT NOT NULL` so not as `VARCHAR`.
Of course there is a case for extracting those to a separate table too - but I don't think we need that for now.

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/random/RandomUtilities.kt
```diff
@@ -31,9 +31,15 @@ internal fun fullyPopulatedRecall(recallId: RecallId = ::RecallId.random()): Rec
...
-          // ensure document version is valid verus category
+          // ensure document version is valid versus category
           version = if (document.category.versioned) Random.nextInt() else null
         )
+      }.toSet(),
+      missingDocumentsRecords = recall.missingDocumentsRecords.map { mdr ->
+        mdr.copy(
+          recallId = recallId.value,
+          emailId = recall.documents.random().id
+        )
```
AndyNewton, line 42: Have you seen any instability in tests using `fullyPopulatedRecall`?   The comment `// ensure document version is valid versus category` and code for that is something I needed to add for that with versioned documents.

AndyNewton (dismissed):
Maybe take a quick look at my latest comments - before or after merge, as you wish.

AndyNewton (approved):

PR #344, opened 2021-11-22
PUD-1030: police force select from list BE step 1

AndyNewton:
PUD-1030: police force select from list BE step 1 (but needs ^^^ PR approved first please).


philsegal (commented):
Cant see the new field being used anywhere? e.g. in the document generator?

AndyNewton:
^^^ the idea here is item 1. from https://dsdmoj.atlassian.net/browse/PUD-453?focusedCommentId=185471, i.e.:
"1. BE: make new values available as ref-data; add the new Id property alongside the String in all APIs, persistence etc.: can be freely promoted."
It's a later step to use the Id in place of the String in the BE.  TBH that does mean that I've included the Id in a couple of classes where it isn't needed or used at this time - e.g. the Contexts.  It'll get used when the String is removed from those classes.

philsegal (approved):

PR #348, opened 2021-11-23
PUD-1039: Clear up unused/old pact states

philsegal:


AndyNewton (approved):
Sorry - got distracted.  
We _could_ tidy some of this wording up a little - but this is fine and good clean-up so 👍 .

PR #364, opened 2021-12-01
Add a script for copying the S3 bucket between envs

dazoakley:


philsegal (approved):
I assume you've run and tested it locally :)

PR #373, opened 2021-12-07
PUD-1131  Remove police force name property entirely from BE

AndyNewton:
PUD-1131  Remove police force name property entirely from BE
Tiny but keeping apart from health-check PR to follow.


philsegal (approved):
Does the db table needs the column dropping?

AndyNewton:
column drop is in this PR @philsegal : see
src/main/resources/db/migration/V63__Remove_localPoliceForce_column.sql

PR #380, opened 2021-12-09
PUD-1071: Clean up GetDocumentResponse. Ensure full name is unwrapped as String in swagger

philsegal:


src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/GetRecallComponentTest.kt
```diff
@@ -132,6 +132,9 @@ class GetRecallComponentTest : ComponentTestBase() {
       .jsonPath("$.dossierTargetDate").isEqualTo(LocalDate.now().toString())
       .jsonPath("$.assignee").isEqualTo(fullyPopulatedRecall.assignee!!.toString())
       .jsonPath("$.assigneeUserName").isEqualTo("Bertie Badger")
+      .jsonPath("$.dossierCreatedByUserName").isEqualTo("Bertie Badger")
```
AndyNewton, line 135: So this is an example of where I think "if only Bertie had a few cousins who also worked for PPCS, how much clearer our tests could be".

AndyNewton (approved):
Approved with 1 predictable comment.

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/GetRecallComponentTest.kt
philsegal: But we'd also need to store a whole load more user details

PR #381, opened 2021-12-09
PUD-1059 Add Welsh recall info leaflet to Dossier for Welsh LDUs

AndyNewton:
PUD-1059 Add Welsh recall info leaflet to Dossier for Welsh LDUs


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/documents/dossier/DossierContextFactory.kt
```diff
@@ -43,6 +43,10 @@ data class DossierContext(
       currentPrisonName,
       recall.bookingNumber!!
     )
+
+  fun includeWelsh(): Boolean {
```
philsegal, line 47: not sure if that's the best name, include suggests there could be others? 'welshOnly'?

philsegal: ah, I see now. `lduIsInWales`?

philsegal (approved):
Small naming comment, else looks good.

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/documents/dossier/DossierContextFactory.kt
AndyNewton: I dunno what would really be better.  Everything else is already included.  This is the line of usage:
`createTableOfContentsDocumentMap(reasonsForRecallPdfBytes, recallId, dossierContext.includeWelsh())`
and the result is to add the Welsh doc.  🤷‍♂️

PR #387, opened 2021-12-13
PUD-1133: Add new endpoint to (re)create generated documents (RECALL_NOTIFICATION only for now)

philsegal:


src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/GetRecallNotificationComponentTest.kt
```diff
@@ -54,6 +55,39 @@ class GetRecallNotificationComponentTest : ComponentTestBase() {
     assertThat(response.content, equalTo(expectedBase64Pdf))
   }
 
+  @Test
+  fun `get recall notification returns merged recall summary and revocation order and then create recall notification creates new version but reuses existing revocation order`() {
```
AndyNewton, line 59: What about this test verifies "but reuses existing revocation order"?

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/DocumentController.kt
```diff
@@ -157,6 +158,35 @@ class DocumentController(
     }
   }
 
+  @PostMapping("/recalls/{recallId}/documents/create")
```
AndyNewton, line 161: Just wondering if we should in due cause remove the create capability from the GET end-points (`/recallNotification` and `/dossier`)?

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/documents/recallnotification/RecallNotificationServiceTest.kt
```diff
@@ -86,6 +88,54 @@ internal class RecallNotificationServiceTest {
     )
   }
 
+  @Test
+  fun `create recall notification uses latest revocation order if exists and creates recall summary and letter to probation`() {
```
AndyNewton, line 92: How about re-wording title as:
"create recall notification uses latest revocation order, or creates if not exists, and creates recall summary and letter to probation"
I think the titles need to make it clearer what each test is verifying please.

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/documents/recallnotification/RecallNotificationServiceTest.kt
```diff
@@ -86,6 +88,54 @@ internal class RecallNotificationServiceTest {
     )
   }
```
AndyNewton, line 90: Can we improve the name of the test before this one?  Currently it reads:
"get recall notification returns recall summary, revocation order and letter to probation when one doesnt exist already".
That doesn't tell me what "one" refers to.  And it's difficult to see how it compares to the following test.

AndyNewton (changes requested):
Per comments, can you see if the commented test titles can be made clearer representations of what each test is supposed to test ... please.

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/GetRecallNotificationComponentTest.kt
philsegal: Well, there are 2 actions here. First one is the initial call to get recall notification, then we call the create endpoint which reuses. so it is wordy, but makes sense

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/DocumentController.kt
philsegal: I'd like to do that :) but would mean hooking the create action into the front end somewhere. One to discuss.

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/DocumentController.kt
```diff
@@ -210,14 +243,19 @@ data class UpdateDocumentResponse(
...
   val details: String? = null
 )
 
-data class AddDocumentResponse(val documentId: DocumentId)
+data class GenerateDocumentRequest(
+  val category: DocumentCategory,
+  val details: String
+)
+
+data class NewDocumentResponse(val documentId: DocumentId)
```
AndyNewton, line 258: Not a fan of this name but we can change later - when we think of a better name.

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/GetRecallNotificationComponentTest.kt
```diff
@@ -49,9 +51,37 @@ class GetRecallNotificationComponentTest : ComponentTestBase() {
...
+    assertThat(getResponse.content, equalTo(expectedBase64Pdf))
+
+    val firstRecallNotification =
+      documentRepository.findLatestVersionedDocumentByRecallIdAndCategory(recall.recallId, RECALL_NOTIFICATION)!!
+    assertThat(firstRecallNotification.version, equalTo(1))
+    val firstRevocationOrder =
+      documentRepository.findLatestVersionedDocumentByRecallIdAndCategory(recall.recallId, REVOCATION_ORDER)!!
+    assertThat(firstRevocationOrder.version, equalTo(1))
+
+    // Reset mocks to ensure that revocation order isnt regenerated
```
AndyNewton, line 65: 👍

AndyNewton (approved):

PR #389, opened 2021-12-13
PUD-1060  Include Welsh recall leaflet for Welsh prisons

AndyNewton:
PUD-1060  Include Welsh recall leaflet for Welsh prisons

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/documents/dossier/DossierContextFactory.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/documents/dossier/DossierContextFactory.kt
@@ -18,14 +18,17 @@ class DossierContextFactory(
 ) {
   fun createContext(recallId: RecallId): DossierContext {
     val recall = recallRepository.getByRecallId(recallId)
-    val currentPrisonName = prisonLookupService.getPrisonName(recall.currentPrison!!)
-    return DossierContext(recall, currentPrisonName)
+    val currentPrisonId = recall.currentPrison!!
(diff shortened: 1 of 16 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/documents/dossier/DossierContextFactory.kt
```diff
@@ -18,14 +18,17 @@ class DossierContextFactory(
 ) {
   fun createContext(recallId: RecallId): DossierContext {
     val recall = recallRepository.getByRecallId(recallId)
-    val currentPrisonName = prisonLookupService.getPrisonName(recall.currentPrison!!)
-    return DossierContext(recall, currentPrisonName)
+    val currentPrisonId = recall.currentPrison!!
+    val currentPrisonName = prisonLookupService.getPrisonName(currentPrisonId)
+    val currentPrisonIsWelsh = prisonLookupService.isWelsh(currentPrisonId)
```
philsegal, line 23: Feels a bit clunky? Would it be better to have a single `prisonLookupService.getPrison` which returns ID, name, active & isWelsh?

philsegal (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/documents/dossier/DossierContextFactory.kt
AndyNewton: I was thinking to change (what is now) the `Api.Prison` class to add an `isWelsh` property and decorating the response object from the Prison register - but currently we only need to know this in one place so I went with putting this only in the LookupService.  I'm fine with looking at the alternative.

PR #403, opened 2021-12-17
PUD-1017: Make details optional on /generated so we can stop using the getOrCreate calls.

philsegal:
Remove unused endpoint.
Prevent upload of generated docs.
Deprecate GET (and/or create) endpoints.


src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/DocumentControllerTest.kt
```diff
@@ -355,4 +369,31 @@ class DocumentControllerTest {
     verify { revocationOrderService wasNot Called }
     verify { reasonsForRecallService wasNot Called }
   }
+
+  @Test
+  fun `generate new document without details`() {
```
AndyNewton, line 331: I think I'd prefer less duplication between these tests and instead re-use of a shared method - but for another time.

AndyNewton (approved):
Approved with 1 comment.

PR #410, opened 2021-12-22
PUD-1237  Update manage-recalls-api to JDK17 (openjdk:17)

AndyNewton:
PUD-1237  Update manage-recalls-api to JDK17 (openjdk:17)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -83,7 +83,7 @@ dependencies {
 tasks {
   withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
     kotlinOptions {
-      jvmTarget = JavaVersion.VERSION_16.toString()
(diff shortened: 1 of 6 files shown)


build.gradle.kts
```diff
@@ -121,7 +121,7 @@ task<Test>("documentGenerationTest") {
 
 java {
   toolchain {
-    languageVersion.set(JavaLanguageVersion.of(16))
+    languageVersion.set(JavaLanguageVersion.of(16)) // TODO Need Kotlin 1.6 to target JVM 17 - see PUD-1241
```
philsegal, line 124: cant we use jdk 17 here, but just accept the warnings?

philsegal (approved):

build.gradle.kts
AndyNewton: As per comment, we'll get there when we go to Kotlin 1.6.

PR #418, opened 2022-01-05
PUD-1250  Change 'detail' property to 'details' in missing documents …

AndyNewton:
PUD-1250  Change 'detail' property to 'details' in missing documents record request and returned object.


AndyNewton:
This branch passes PACT vs UI main and in local testing also passes PACT vs UI branch `PUD-1073-missing-doc-records`.

philsegal (approved):
Ew, breaking changes! But I can see why you've done what you've done!

PR #422, opened 2022-01-07
PUD-1259  missing documents records BE: remove transitional API code

AndyNewton:
PUD-1259  missing documents records BE: remove transitional API code now that UI change is complete
FYI I'm still looking at improving the PACT in this area, e.g. given that the UI contract "expected" `emailFileName` in the MDR when it for certain was not present.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/MissingDocumentsRecordComponentTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/MissingDocumentsRecordComponentTest.kt
@@ -9,7 +9,7 @@ import org.springframework.http.HttpStatus.BAD_REQUEST
 import org.springframework.http.HttpStatus.CREATED
 import uk.gov.justice.digital.hmpps.managerecallsapi.config.ErrorResponse
 import uk.gov.justice.digital.hmpps.managerecallsapi.controller.BookRecallRequest
-import uk.gov.justice.digital.hmpps.managerecallsapi.controller.TempMissingDocumentsRecordRequest
+import uk.gov.justice.digital.hmpps.managerecallsapi.controller.MissingDocumentsRecordRequest
 import uk.gov.justice.digital.hmpps.managerecallsapi.db.DocumentCategory.PART_A_RECALL_REPORT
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/MissingDocumentsRecordComponentTest.kt
```diff
@@ -50,59 +49,40 @@ class MissingDocumentsRecordComponentTest : ComponentTestBase() {
...
+    val detailsMostRecent = "Some details; some more detail"
+    val response = authenticatedClient.missingDocumentsRecord(missingDocsRecordReq.copy(details = detailsMostRecent), CREATED, MissingDocumentsRecordId::class.java)
 
     assertThat(response, present())
 
     val recallWithMissingDocumentsRecord = authenticatedClient.getRecall(recall.recallId)
     assertThat(recall.missingDocumentsRecords, isEmpty)
     assertThat(recallWithMissingDocumentsRecord.missingDocumentsRecords.size, equalTo(2))
     assertThat(recallWithMissingDocumentsRecord.missingDocumentsRecords[0].version, !equalTo(recallWithMissingDocumentsRecord.missingDocumentsRecords[1].version))
+    // TODO: following are unstable with the order of the returned records being non-deterministic when presumably we do want them to be
```
AndyNewton, line 75: This surprised me: and I didn't yet find why it might be this way.

philsegal: that's odd that its seemingly been working...

philsegal (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/MissingDocumentsRecordComponentTest.kt
AndyNewton: I suspect it has not been working: the commented asserts are new here too.  Just we should be aware of this - which perhaps we weren't before? - and preferably fix it.

PR #428, opened 2022-01-12
PUD-1201: Fix failure with nested properties blowing up

philsegal:
- flatten FieldName in json


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallAuditService.kt
```diff
@@ -88,11 +91,26 @@ class RecallAuditService(
...
   }
 
-  private fun getFieldNameForColumn(columnName: ColumnName): FieldName {
-    return Recall::class.java.declaredFields.first { field ->
-      getColumnNameForField(FieldName(field.name)) == columnName
-    }.let { FieldName(it.name) }
-  }
+  private fun getFieldNameForColumn(columnName: ColumnName): FieldName =
+    Recall::class.java.declaredFields.flatMap { field ->
+      val propCols = persister.getPropertyColumnNames(field.name).map { ColumnName(it) to FieldName(field.name) }
```
AndyNewton, line 96: 🤔  Without trying it ... feels like this might be simplified by use of recursion?

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/RecallAuditComponentTest.kt
```diff
@@ -143,6 +143,11 @@ class RecallAuditComponentTest : ComponentTestBase() {
     assertThat(updatedAuditList[3].updatedByUserName, equalTo(FullName("Bertie Badger")))
     assertThat(updatedAuditList[3].auditCount, equalTo(2))
     assertOffsetDateTimesEqual(updatedAuditList[3].updatedDateTime, updatedRecall.lastUpdatedDateTime)
+
+    // assert that still works with sentencing info added
```
AndyNewton, line 147: Can the comment say why sentencing info is special/this test is needed please?

AndyNewton (dismissed):
Couple of QQs - for next PR?

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallAuditService.kt
philsegal: Yes, possibly. Head was starting to hurt. Def room for optimisation, however we might need to restrict depth of recursion.

AndyNewton (approved):

PR #431, opened 2022-01-14
PUD-1227: Use get /prisoner rather than search when getting a single prisoner by noms number

philsegal:


src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/pact/provider/ManageRecallsUIPactTest.kt
```diff
@@ -84,10 +84,10 @@ class ManagerRecallsUiAuthorizedPactTest : ManagerRecallsUiPactTestBase() {
 
   @State("a prisoner exists for NOMS number")
   fun `a prisoner exists for NOMS number`() {
-    mockPrisonerResponse(nomsNumber)
+    mockPrisonerResponses(nomsNumber)
```
AndyNewton, line 87: So we're not adding to the contract for the return of a single prisoner outside of a List/Array?

AndyNewton (approved):
QQ what the plan is for the API to evolve; not changed here I think?

PR #433, opened 2022-01-17
PUD-1282  Fix up existing data prior to constraint application

AndyNewton:
Should fix up existing data for included new DB constraints as per promote issue per Slack: https://mojdt.slack.com/archives/C0223AGGQU8/p1642415970177700?thread_ts=1642414658.172800&cid=C0223AGGQU8

### DIFF
--- a/src/main/resources/db/migration/V68__Non_blank_details_for_document_version_gt_one.sql
+++ b/src/main/resources/db/migration/V68__Non_blank_details_for_document_version_gt_one.sql
@@ -1,10 +1,21 @@
--- Allow blank document details only when version is null or 1
-ALTER TABLE document ADD CONSTRAINT details_not_blank_constraint_for_version_gt_1
-    CHECK ((version IS NULL) OR
-           (version = 1) OR
-           (details IS NOT NULL AND TRIM(details) <> ''));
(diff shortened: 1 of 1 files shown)


src/main/resources/db/migration/V68__Non_blank_details_for_document_version_gt_one.sql
```diff
@@ -1,10 +1,21 @@
--- Allow blank document details only when version is null or 1
-ALTER TABLE document ADD CONSTRAINT details_not_blank_constraint_for_version_gt_1
-    CHECK ((version IS NULL) OR
-           (version = 1) OR
-           (details IS NOT NULL AND TRIM(details) <> ''));
+-- Patch existing data prior to constraint: Restrict valid values of version: not _expected_ to apply to any data
+UPDATE document SET version = 9999
+WHERE version IS NOT NULL AND
+        version < 1;
```
philsegal, line 4: 🤯

philsegal (approved):

src/main/resources/db/migration/V68__Non_blank_details_for_document_version_gt_one.sql
AndyNewton: This is for the constraint which I don't expect to be broken: yes it's ugly; but it has to do something with any such data.

PR #434, opened 2022-01-18
Add encrypted secrets

dazoakley:
This encoded the manually created secrets within the codebase using `git-crypt`.

ref https://www.agwa.name/projects/git-crypt/
ref PUD-22


dazoakley:
Turns out these secrets are automated now, so no need to do this here: https://mojdt.slack.com/archives/CR5ESQ8T1/p1642671807043500?thread_ts=1642602870.034400&cid=CR5ESQ8T1

PR #439, opened 2022-01-21
PUD-136: Add last known address option field & PUD-634: Add Capture Arrest Issues fields

philsegal:


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallController.kt
```diff
@@ -458,3 +467,8 @@ enum class NameFormatCategory {
   FIRST_MIDDLE_LAST,
   OTHER
 }
+
+enum class LastKnownAddressCategory {
```
AndyNewton: Just in case we want to change, per Slack chat https://mojdt.slack.com/archives/C0223AGGQU8/p1642778612324600?thread_ts=1642774696.318500&cid=C0223AGGQU8

AndyNewton (approved):

PR #443, opened 2022-01-25
Let the user know how to install Java, to run API locally

jonnywyatt:


dazoakley (approved):

.tool-versions
```diff
@@ -0,0 +1 @@
+java openjdk-17.0.2
```
philsegal, line 1: do we want to commit this? We're not all using ASDF

philsegal (changes requested):
Not sure this is needed. Superseded by https://github.com/ministryofjustice/manage-recalls-api/pull/444 which will install the correct version automagically.

PR #445, opened 2022-01-25
PUD-1281: Delete a last known address.

philsegal:
- add support for /recalls/{recallId}/* for MissingDocumentsRecords & LastKnownAddresses

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/LastKnownAddressController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/LastKnownAddressController.kt
@@ -9,10 +9,13 @@ import org.springframework.http.HttpStatus
 import org.springframework.http.MediaType
 import org.springframework.http.ResponseEntity
 import org.springframework.security.access.prepost.PreAuthorize
+import org.springframework.web.bind.annotation.DeleteMapping
+import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestHeader
 import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.ResponseStatus
 import org.springframework.web.bind.annotation.RestController
 import uk.gov.justice.digital.hmpps.managerecallsapi.config.ErrorResponse
 import uk.gov.justice.digital.hmpps.managerecallsapi.controller.extractor.TokenExtractor
@@ -23,6 +26,7 @@ import uk.gov.justice.digital.hmpps.managerecallsapi.db.RecallRepository
 import uk.gov.justice.digital.hmpps.managerecallsapi.domain.LastKnownAddressId
 import uk.gov.justice.digital.hmpps.managerecallsapi.domain.RecallId
 import uk.gov.justice.digital.hmpps.managerecallsapi.domain.random
+import uk.gov.justice.digital.hmpps.managerecallsapi.service.LastKnownAddressNotFoundException
 import java.time.OffsetDateTime
 
 @RestController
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/LastKnownAddressController.kt
```diff
@@ -37,21 +41,31 @@ class LastKnownAddressController(
...
+        ExampleProperty(
+          mediaType = "application/json",
+          value = "{\n\"status\": 404,\n\"message\":\"RecallNotFoundException(recallId=...)\"\n}"
+        )
+      )
     )
   )
-  @PostMapping("/last-known-addresses")
+  @PostMapping(
+    "/recalls/{recallId}/last-known-addresses",
```
AndyNewton, line 53: OK; as per thread, not sure we need this but 🤷‍♂️

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/LastKnownAddressComponentTest.kt
```diff
@@ -109,9 +111,33 @@ class LastKnownAddressComponentTest : ComponentTestBase() {
...
     val expectedStatus = HttpStatus.NOT_FOUND
-    val response = authenticatedClient.lastKnownAddress(createLastKnownAddressRequest, expectedStatus, ErrorResponse::class.java)
+    val response = authenticatedClient.addLastKnownAddress(notFoundRecallId, createLastKnownAddressRequest, expectedStatus, ErrorResponse::class.java)
 
     assertThat(response, present())
     assertThat(response, equalTo(ErrorResponse(expectedStatus, "RecallNotFoundException(recallId=$notFoundRecallId)")))
   }
+
+  @Test
+  @Transactional
```
AndyNewton, line 121: We're not using `@Transactional` on any other component tests that I can see ... so why here?

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/integration/db/LastKnownAddressRepositoryIntegrationTest.kt
```diff
@@ -103,4 +105,37 @@ class LastKnownAddressRepositoryIntegrationTest : IntegrationTestBase() {
...
     OffsetDateTime.now()
   )
+
+  @Test
+  @Transactional
+  fun `deleting a last known address that exists for recall`() {
+    val lastKnownAddressId = ::LastKnownAddressId.random()
+    recallRepository.save(recall, currentUserId)
+
+    lastKnownAddressRepository.saveAndFlush(lastKnownAddress(lastKnownAddressId, 1))
```
AndyNewton, line 115: Just 🤔  whether `saveAndFlush` allows for the retrieve via the recall to work in these tests too?
And whether we should move to `saveAndFlush` in place of `save` for all tests in this directory - with @Transactional on them all too, I'd assume.

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/integration/db/LastKnownAddressRepositoryIntegrationTest.kt
```diff
@@ -103,4 +105,37 @@ class LastKnownAddressRepositoryIntegrationTest : IntegrationTestBase() {
...
+      lastKnownAddressRepository.deleteByRecallIdAndLastKnownAddressId(recall.recallId(), lastKnownAddressId)
+    }
+  }
+
+  @Test
+  @Transactional
+  fun `deleting a last known address that exists for different recall throws exception`() {
+    val lastKnownAddressId = ::LastKnownAddressId.random()
+    recallRepository.save(recall, currentUserId)
+    lastKnownAddressRepository.save(lastKnownAddress(lastKnownAddressId, 1))
```
AndyNewton, line 135: Happy case test uses `lastKnownAddressRepository.saveAndFlush` so preferable here too IMO.

AndyNewton (changes requested):
Couple of test tweaks please.

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/integration/db/LastKnownAddressRepositoryIntegrationTest.kt
philsegal: nope. It didn't make a difference. The only place it seems to is when we create recalls with documents.

AndyNewton (approved):

PR #464, opened 2022-02-02
Use a specific release of minio

dazoakley:
Using `latest` probably isn't the best thing (as it's a moving target), so let's see if this resolves the CircleCI issues...


dazoakley:
I've removed the requirement on the `trivy_scan` so this can get in, then @philsegal can add his fix for that CVE after.

philsegal (approved):

PR #467, opened 2022-02-02
PUD-1427  BE: Add new Recall status AWAITING_RETURN_TO_CUSTODY

AndyNewton:
PUD-710 & PUD-717: NIC assessment: Letter to probation variation


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallController.kt
```diff
@@ -401,6 +401,7 @@ enum class Status(val visibilityBands: Set<CaseworkerBand>) {
   BOOKED_ON(FOUR_PLUS_ONLY),
   IN_ASSESSMENT(FOUR_PLUS_ONLY),
   RECALL_NOTIFICATION_ISSUED(ALL_BANDINGS),
+  AWAITING_RETURN_TO_CUSTODY(ALL_BANDINGS),
```
philsegal, line 404: Have we confirmed this is `ALL_BANDINGS`?

philsegal (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallController.kt
AndyNewton: This just sets UI list visibility right?  I don't recall any mention of bands.  How about we merge as is and I ping product a note on Slack?

PR #468, opened 2022-02-03
PUD-1433  Back End - Step 1: add two new Recall properties

AndyNewton:
PUD-1433  Back End - Step 1: add two new Recall properties: inCustodyAtBooking and inCustodyAtAssessment
Leave old property unchanged.  Include some usage adjustments to account for old and new, e.g. for document generation.
Passes e2es locally.


philsegal (dismissed):

philsegal (approved):

src/main/resources/db/migration/V77__Add_two_in_custody_booleans.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE recall
+    ADD COLUMN in_custody_at_booking BOOLEAN default NULL,
+    ADD COLUMN in_custody_at_assessment BOOLEAN default NULL;
+UPDATE recall SET in_custody_at_booking = in_custody;
+UPDATE recall SET in_custody_at_assessment = in_custody;
```
philsegal, line 5: argh. not sure this should be set if in_custody = true

PR #499, opened 2022-02-21
Allow network ingress from the `manage-recalls-dev` namespace

dazoakley:
This is so that the dashboards app can get the current version of the
application - this is now only available internally to the cluster.


philsegal (approved):
I trust you

dazoakley:
Thanks! 😃

PR #512, opened 2022-02-24
PUD-1512: Allow replacement of emails without needing to supply details

philsegal:


src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/DocumentServiceTest.kt
```diff
@@ -163,8 +163,9 @@ internal class DocumentServiceTest {
     every { s3Service.uploadFile(capture(uploadedToS3DocumentIdSlot), documentBytes) } just runs
     every { documentRepository.save(capture(savedDocumentSlot)) } returns document
 
+    val details = "Some details"
     val actualDocumentId =
-      underTest.storeDocument(recallId, createdByUserId, documentBytes, documentCategory, newFileName)
+      underTest.storeDocument(recallId, createdByUserId, documentBytes, documentCategory, newFileName, details)
```
AndyNewton, line 168: Not this change but less than obvious that `documentCategory` is a reference to a versioned category:
`private val documentCategory = PART_A_RECALL_REPORT`
Maybe `val documentCategory` local to each test is better than how it is in this class now?

AndyNewton (changes requested):
Take a look and give me a call when you have the chance.

AndyNewton (approved):

PR #515, opened 2022-02-28
PUD-1517: Add metrics for automated return to custody.

philsegal:


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallService.kt
```diff
@@ -145,7 +150,10 @@ class RecallService(
         prisonApiClient.latestInboundMovements(rtcRecalls.map { it.nomsNumber }.toSet()).associateBy { it.nomsNumber() }
 
       rtcRecalls.forEach {
-        returnedToCustody(it, movements[it.nomsNumber]!!.movementDateTime(), OffsetDateTime.now(clock), SYSTEM_USER_ID)
+        val returnedToCustodyDateTime = movements[it.nomsNumber]!!.movementDateTime()
+        log.info("Returning ${it.recallId()} to custody as of $returnedToCustodyDateTime")
+        returnedToCustody(it, returnedToCustodyDateTime, OffsetDateTime.now(clock), SYSTEM_USER_ID)
+        meterRegistry.counter("autoReturnedToCustody")
```
dazoakley, line 156: Does this actually increment the counter?

philsegal: `Tracks a monotonically increasing value.
Params:
name – The base metric name
tags – MUST be an even number of arguments representing key/value pairs of tags.
Returns:
A new or existing counter.`
Looks like it...

dazoakley: Pretty sure you'll need to call `.increment()`

dazoakley (approved):

PR #528, opened 2022-03-02
Initialise the rtcCounter on start

dazoakley:
The lazy initialisation of the counter wasn't giving us a default zero
counter on application load, this change fixes that.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallServiceTest.kt
@@ -9,6 +9,7 @@ import io.mockk.every
 import io.mockk.just
 import io.mockk.mockk
 import io.mockk.verify
+import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertThrows
 import reactor.core.publisher.Mono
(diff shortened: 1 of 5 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallServiceTest.kt
```diff
@@ -54,11 +55,12 @@ class RecallServiceTest {
...
   private val prisonApiClient = mockk<PrisonApiClient>()
-  private val meterRegistry = mockk<MeterRegistry>()
+  private val meterRegistry = mockk<MeterRegistry>(relaxed = true)
+  private val autoReturnedToCustodyCounter = mockk<Counter>()
   private val fixedClock = Clock.fixed(Instant.parse("2021-10-04T13:15:50.00Z"), ZoneId.of("UTC"))
   private val returnToCustodyUpdateThresholdMinutes = 60L
-  private val underTest =
-    RecallService(recallRepository, bankHolidayService, prisonerOffenderSearchClient, prisonApiClient, fixedClock, meterRegistry, returnToCustodyUpdateThresholdMinutes)
+
+  private var underTest = mockk<RecallService>()
```
philsegal, line 63: not sure why we're setting it as a `mock<RecallService>()`. It def shouldn't ever be a mock

dazoakley: I tried `null` but I couldn't then re-assign it as a `RecallService` in the `@BeforeEach` block...

I could just assign it as before here and then re-assign after the two `every` blocks in the `@BeforeEach`, happy to do whatever...

philsegal: `private lateinit var underTest: RecallService`

philsegal (approved):

PR #541, opened 2022-03-11
Add timeout counters for the "search" clients

dazoakley:
I missed these when doing the registers the other day... There are no (unit) tests on these clients, I'm happy to add some, but not really sure where to start - anyone for pairing?

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -31,6 +31,3 @@ generic-service:
 
 generic-prometheus-alerts:
   alertSeverity: ppud-replacement-alerts
-  extraDashboardTags:
-    - manage-recalls
-    - dev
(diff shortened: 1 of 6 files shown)


helm_deploy/values-dev.yaml
```diff
@@ -31,6 +31,3 @@ generic-service:
 
 generic-prometheus-alerts:
   alertSeverity: ppud-replacement-alerts
-  extraDashboardTags:
```
AndyNewton, line 34: Unrelated clean-up?

AndyNewton (changes requested):
I see no tests for the added funtionality @dazoakley .  Can we have some please!  Thanks.

helm_deploy/values-dev.yaml
dazoakley: Yes, seemed pointless having a separate PR for it...

dazoakley:
> I see no tests for the added funtionality @dazoakley . Can we have some please! Thanks.

So is that an offer to pair/help writing them then?

AndyNewton (approved):

PR #545, opened 2022-03-14
It looks like I labelled ClientTimeoutException wrong...

dazoakley:
Sentry is still reporting errors.

### DIFF
--- a/helm_deploy/manage-recalls-api/values.yaml
+++ b/helm_deploy/manage-recalls-api/values.yaml
@@ -33,7 +33,7 @@ generic-service:
     SPRING_PROFILES_ACTIVE: "logstash"
     APPLICATIONINSIGHTS_CONNECTION_STRING: "InstrumentationKey=$(APPINSIGHTS_INSTRUMENTATIONKEY)"
     GOTENBERG_ENDPOINT_URL: http://manage-recalls-api-gotenberg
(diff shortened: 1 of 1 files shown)


helm_deploy/manage-recalls-api/values.yaml
```diff
@@ -33,7 +33,7 @@ generic-service:
     SPRING_PROFILES_ACTIVE: "logstash"
     APPLICATIONINSIGHTS_CONNECTION_STRING: "InstrumentationKey=$(APPINSIGHTS_INSTRUMENTATIONKEY)"
     GOTENBERG_ENDPOINT_URL: http://manage-recalls-api-gotenberg
-    SENTRY_IGNORED_EXCEPTIONS_FOR_TYPE: uk.gov.justice.digital.hmpps.managerecallsapi.service.VirusFoundException,org.springframework.security.web.firewall.RequestRejectedException,org.springframework.web.HttpMediaTypeNotAcceptableException,uk.gov.justice.digital.hmpps.managerecallsapi.config.ClientTimeoutException
+    SENTRY_IGNORED_EXCEPTIONS_FOR_TYPE: uk.gov.justice.digital.hmpps.managerecallsapi.service.VirusFoundException,org.springframework.security.web.firewall.RequestRejectedException,org.springframework.web.HttpMediaTypeNotAcceptableException,ClientTimeoutException
```
AndyNewton, line 36: QQ: Apparently this, the first entry in the list, is right:
`uk.gov.justice.digital.hmpps.managerecallsapi.service.VirusFoundException`
so I don't see why the unqualified class name would be right for `ClientTimeoutException`?
The code also shows the fully-qualified previous value (`uk.gov.justice.digital.hmpps.managerecallsapi.config.ClientTimeoutException`) as correct.  🤷‍♂️

dazoakley: It's probably some Mono/Webclient BS as it's titled as a `Exceptions$ReactiveException`, so I could just ignore that but I'm guessing that'll swallow a lot of things we want to surface.

PR #568, opened 2022-03-25
PUD-1632: Correct letters to match existing documents

philsegal:


src/test/resources/uk/gov/justice/digital/hmpps/managerecallsapi/documents/lettertoprison/LetterToPrisonGovernorHtmlGenerationTest_generate fully populated HTML FTR 14 days.approved
```diff
@@ -369,9 +369,9 @@ line-height:normal'>&nbsp;</p>
 line-height:normal'><span dir=LTR></span><b>Any representations must be scanned and e-mailed
     to the Public Protection Casework Section, Public Protection Group, no later
     than
-<span><span>14</span> calendar </span>
+<span>14</span>
```
dazoakley, line 372: Any particular reason for the double `<span>` in all of these templates?  I mean the browser will just show them on the same line and strip the whitespace anyhow, but seeing this in the code sets my OCD off. 😁

dazoakley (approved):

src/test/resources/uk/gov/justice/digital/hmpps/managerecallsapi/documents/lettertoprison/LetterToPrisonGovernorHtmlGenerationTest_generate fully populated HTML FTR 14 days.approved
philsegal: The old template had double span, its to do with the way that thymeleaf handles optionality. So the first span was the if/then span, and the inner span is inserting the value.

PR #574, opened 2022-03-31
PUD-1645: Endpoint to start/end phase. Start 'BOOK' phase as part of booking a recall.

philsegal:


src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/AuthenticatedClient.kt
```diff
@@ -188,6 +192,19 @@ class AuthenticatedClient(
     sendPostRequest("/recalls/$recallId/stop", StopRecallRequest(stopReason))
       .expectStatus().isEqualTo(OK)
 
+  fun startPhase(recallId: RecallId, phase: Phase, expectedStatus: HttpStatus = CREATED) =
```
AndyNewton, line 195: We don't need to change anything right now but FYI I do wonder why we have e.g. two methods for `startPhase` where one does not call the other and hence they duplicate the URL.  Then the `endPhase` seems like another model again - where e.g. the expectBody is explicit not a parameter to a generic `patchRequest` method? 🤷‍♂️

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/BookRecallComponentTest.kt
```diff
@@ -44,6 +46,16 @@ class BookRecallComponentTest : ComponentTestBase() {
         has(RecallResponse::licenceNameCategory, equalTo(NameFormatCategory.FIRST_LAST)),
       )
     )
+
+    val phaseRecords = phaseRecordRepository.findAll()
+    println(phaseRecords[0])
```
AndyNewton, line 51: `println(phaseRecords[0])` ?

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/PhaseRecordComponentTest.kt
```diff
@@ -0,0 +1,116 @@
...
+
+  @Test
+  fun `start the BOOK Phase for recall returns BAD_REQUEST`() {
+    val recall = authenticatedClient.bookRecall(bookRecallRequest)
+
+    authenticatedClient.startPhase(recall.recallId, Phase.BOOK, BAD_REQUEST)
+  }
+
+  @Test
+  fun `start the Phase for recall creates new record`() {
```
AndyNewton, line 35: Maybe: "starting a Phase (other than BOOK) for recall ..."

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/PhaseRecordComponentTest.kt
```diff
@@ -0,0 +1,116 @@
...
+    assertThat(record.recallId(), equalTo(recall.recallId))
+    assertThat(record.phase, equalTo(phase))
+    assertThat(record.startedByUserId(), equalTo(authenticatedClient.userId))
+    assertThat(record.startedDateTime, !equalTo<OffsetDateTime>(null))
+    assertThat(record.endedByUserId, equalTo(null))
+    assertThat(record.endedDateTime, equalTo(null))
+  }
+
+  @Test
+  fun `starting the phase twice for a recall overwrites first record`() {
```
AndyNewton, line 50: "starting the phase" makes me expect the phase to be identified; maybe "the" -> "a" for these test titles?

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/PhaseRecordComponentTest.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    assertThat(endRecord.id(), equalTo(startRecord.id()))
+    assertThat(endRecord.phase, equalTo(startRecord.phase))
+    assertThat(endRecord.phase, equalTo(phase))
+    assertOffsetDateTimesEqual(endRecord.startedDateTime, startRecord.startedDateTime)
+    assertThat(endRecord.startedByUserId(), equalTo(startRecord.startedByUserId()))
+    assertThat(endRecord.endedDateTime!!.isAfter(endRecord.startedDateTime), equalTo(true))
+    assertThat(endRecord.endedByUserId(), equalTo(authenticatedClient.userId))
+  }
+
```
AndyNewton, line 89: What about when we try to end a phase that hasn't been started?

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/RecallControllerTest.kt
```diff
@@ -112,23 +111,13 @@ class RecallControllerTest {
   fun `book recall returns request with id`() {
     val bearerToken = "Bearer header.payload"
     val userUuid = ::UserId.random()
-    val recall = recallRequest.toRecall(userUuid, fixedClock)
```
AndyNewton, line 115: `fixedClock` is now unused; should be deleted.

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RecallServiceTest.kt
```diff
@@ -565,4 +569,28 @@ class RecallServiceTest {
     verify { recallRepository.getByRecallId(recallId) }
     verify { recallRepository.save(updatedRecall, currentUserId) }
   }
+
+  @Test
+  fun `book recall also begins phase for BOOK`() {
```
AndyNewton, line 574: "also begins phase for BOOK" -> "starts BOOK phase"?

AndyNewton (changes requested):
Few suggestions

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/PhaseController.kt
```diff
@@ -0,0 +1,76 @@
...
+  @PatchMapping("/recalls/{recallId}/end-phase")
+  @ResponseStatus(HttpStatus.OK)
+  @Transactional
+  fun endPhase(
+    @PathVariable("recallId") recallId: RecallId,
+    @RequestBody request: EndPhaseRequest,
+    @RequestHeader("Authorization") bearerToken: String
+  ): PhaseRecord =
+    tokenExtractor.getTokenFromHeader(bearerToken).userUuid().let { currentUserId ->
+      if (request.shouldUnassign) {
```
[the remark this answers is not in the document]
philsegal: I tried that, but you get cyclical dependencies between RecallService & PhaseService

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/AuthenticatedClient.kt
philsegal: the 2 methods is something we have in lots of places. Where we dont return an object (e.g. it errors) we need to have 1 method, and then the 2nd translates the response into the class.
As for the patch, we dont return anything on the other patch calls, so needed to add it here. 

Possibly we can look at rewriting all the AuthClient at some point, but there always seems to be loads of different permutations. maybe a smell that we aren't consistent in our responses/behaviours

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/PhaseRecordComponentTest.kt
philsegal: ooh, good spot, will add

AndyNewton (approved):
Approved ... with provisos per Slack 😁

PR #578, opened 2022-04-01
PUD-1679 Slice 1 of remove <ResponseEntity<T>> return types

AndyNewton:
PUD-1679 Slice 1 of remove <ResponseEntity<T>> return types in Controllers
Code clean-up: let's agree if this is good before doing much more of the same.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/DocumentController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/DocumentController.kt
@@ -75,20 +75,18 @@ class DocumentController(
   fun getRecallDocument(
     @PathVariable("recallId") recallId: RecallId,
     @PathVariable("documentId") documentId: DocumentId
-  ): ResponseEntity<GetDocumentResponse> {
+  ): GetDocumentResponse {
     val (document, bytes) = documentService.getDocument(recallId, documentId)
 
-    return ResponseEntity.ok(
-      GetDocumentResponse(
-        documentId,
-        document.category,
-        bytes.encodeToBase64String(),
-        document.fileName,
-        document.version,
-        document.details,
-        userDetailsService.get(document.createdByUserId()).fullName(),
-        document.createdDateTime
-      )
+    return GetDocumentResponse(
+      documentId,
+      document.category,
+      bytes.encodeToBase64String(),
+      document.fileName,
+      document.version,
+      document.details,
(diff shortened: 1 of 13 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/DocumentController.kt
```diff
@@ -75,42 +75,37 @@ class DocumentController(
   fun getRecallDocument(
```
philsegal, line 75: Does it not need a `@ResponseStatus(HttpStatus.OK)` ?

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/component/RescindRecordComponentTest.kt
```diff
@@ -148,15 +153,15 @@ class RescindRecordComponentTest : ComponentTestBase() {
...
+    val decisionRequest = RescindRecordController.RescindDecisionRequest(
       true,
       "Some details",
       LocalDate.now(),
       base64EncodedDocumentContents,
       approvalFileName
     )
 
-    authenticatedClient.decideRescind(recall.recallId, ::RescindRecordId.random(), putRequest, HttpStatus.NOT_FOUND)
+    decideRescindFailure(recall, ::RescindRecordId.random(), decisionRequest, NOT_FOUND)
```
philsegal, line 164: Since we're now getting the ErrorResponse class, we could now assert on the error message

src/test/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/service/RescindRecordServiceTest.kt
```diff
@@ -305,7 +305,7 @@ class RescindRecordServiceTest {
   }
 
   @Test
-  fun `cant update a record which has already been decided`() {
+  fun `can not update a record which has already been decided`() {
```
philsegal, line 308: No way 🤣

src/main/kotlin/uk/gov/justice/digital/hmpps/managerecallsapi/controller/DocumentController.kt
AndyNewton: No because that's the default.  And the returned 200/OK for success is tested by the component test BUT it is hidden beyond the `authenticatedClient.getDocument` calls.

philsegal (approved):

PR #586, opened 2022-04-14
Pud 1704

AndyNewton:


scripts/port-forward-db.sh
```diff
@@ -48,7 +48,7 @@ done
 
 # check for the ENV variable
 set +u
-if [[ ! "${ENV}" =~ ^(dev|preprod)$ ]]; then
+if [[ ! "${ENV}" =~ ^(dev|preprod|prod)$ ]]; then
```
philsegal, line 51: do we want to commit this?

AndyNewton: Why would we not?

PR #588, opened 2022-04-21
PUD-1713: Add swagger descriptions

philsegal:


AndyNewton (approved):
Applied a couple of text fixes back to the branch.

